PR #5, opened 2014-11-21
Jhford rewrite

jhford:
I have the basic API written.  I'm not sure how to get generate an api reference json blob, so i don't have it in the client, which results in Auth being next to impossible to do.  For now, I have the deferred authentication commented out.

All of the messages (other than ping) have JSON Schema set up, and I don't suspect that they will change.  I need to add a stats method as well.

I have a couple of questions in routes/v1.js regarding scopes, and I'm not sure what you mean about the scope escalation thing.

For the Worker Type entity, I'm storing things in entity keys, using JSON to represent them, which is working well for all cases other than when a value is undefined.  It would be nice if we could store an entity value as undefined.


jhford:
This is old now, opening a new pr instead

PR #6, opened 2015-02-13
Use Joi to validate provisioner config

rail:
... just to use the same pattern as other taskcluster modules. Also it looks more compact. :)
+ some fixes


rail:
Looks like it's bitrotten again. I'll wait until the transition settles down and will submit a PR if needed.

jhford:
Sorry, I really hate to do this, but there is a bunch of churn on the code
right now.  I'm also thinking about using this to validate incoming
parameters to all functions if it's lightweight enough... Do you have info
on overhead?

rail:
No worries. That was just wrong timing from my side. Whenever you are done with this rewrite, I'd be happy to poke around again.

PR #7, opened 2015-03-02
First overall review

jhford:
Hey @jonasfj, can you take a look at the provisioner.

There are four major moving parts in the provisioner:

* Provisioner
* AwsState
* AwsPricing
* WorkerType

The Provisioner knows how to coordinate between state, pricing and workerTypes.  AwsState knows how to retreive and inspect aws state.  AwsPricing knows how to retreive and inspect AwsPricing.  WorkerType knows how to do things related to a specific workerType, including actually turning their lights on.

Can you take a high level look and let me know if you have any concerns?


TODO.md
```diff
@@ -0,0 +1,26 @@
+## TODO
+
+* Add datapoints in provisioning logic for Influx
+* Be more specific in JSON Schema
+* Load things from NextToken when retreiving spot price history
+* Overwrite UserData with temporary taskcluster credentials
+  * Include the API endpoint for the slave to hit the hello
+    endpoint
+  * include the instanceId scope!
```
jonasfj: Isn't it the `assume:worker-id:<workerGroup>:<workerId>` scope, of which `capacity` number must be issued...
(This need to be coordinated with docker-worker -- let's discuss setting up a meeting).

TODO.md
```diff
@@ -0,0 +1,26 @@
...
+  * cancelling a spot request
+  * creating a spot request
+* Use Joi/assert to better control the input into functions
+* Verify that errors in provisioning don't kill the entire
+  provisioner
+* Kill capacity which exceeds maxCapacity
+* Provide facilities for copying AMIs between regions
+* Api endpoints for
+  * cancelling spot requests
+  * checking on status of spot request
```
jonasfj: Please draft a bug on these... what is the use-case?

TODO.md
```diff
@@ -0,0 +1,26 @@
...
+  * metrics on current state
+* insert into influx when the hello endpoint is hit
```
jonasfj: Not sure what you're thinking...

TODO.md
```diff
@@ -0,0 +1,26 @@
...
+* Load things from NextToken when retreiving spot price history
+* Overwrite UserData with temporary taskcluster credentials
+  * Include the API endpoint for the slave to hit the hello
+    endpoint
+  * include the instanceId scope!
+* Encrypt UserData using opengpg.js
+* Send pulse messages for
+  * killing an instance
+  * cancelling a spot request
+  * creating a spot request
```
jonasfj: Maybe, just with API end-points your todo says to make I don't see the use-case for anybody but the provisioner logic to mess with spot requests...

In the [original design](https://etherpad.mozilla.org/jonasfj-taskcluster-aws-provisioner-strategies) we discussed following exchanges:
```
worker-type-added    (message is status)
worker-type-updated  (message is status)
worker-type-removed  (message is status)
state-updated (message is #capacity, #pending, #nodes, #running)
```
We obviously can't post the worker definition with `worker-type-*` messages, but we can post the state as best we know it. The idea was that the web nodes could store the state from `state-updated` in memory and return it whenever someone asks for the state by API, then people can either poll for the state of listen on pulse.
(State is naturally published after each provisioning iteration when the state has been updated).

TODO.md
```diff
@@ -0,0 +1,26 @@
...
+* Use Joi/assert to better control the input into functions
+* Verify that errors in provisioning don't kill the entire
+  provisioner
+* Kill capacity which exceeds maxCapacity
```
jonasfj: This is blocking. If we're not killing nodes if we launch too many.. (or too many just shows up for no reason). We have problem. We should probably log cases where we kill nodes because we're over the total max. It'll happen occasionally because ec2 is eventually consistent, but it shouldn't happen often.

We need it as a last guard against bad things (tm).

bin/manage-worker-types.js
```diff
@@ -0,0 +1,141 @@
+'use strict';
+process.env.DEBUG='';
+
```
jonasfj: You should really write some comments..
For bin things that a non-obvious it certainly makes sense at at least add... something like, `/** this utility helps you manage.... .. .. . .. . */`

bin/manageTable.js
```diff
@@ -0,0 +1,59 @@
...
+      console.error('Other error while creating table!', x);
+    }
+  }).done();
+} else if (action === 'delete') {
+  WorkerType.removeTable().then(function(x) {
+    console.log('Table Deleted'); 
+  }).catch(function(x) {
+    if (x.statusCode === 404) {
+      console.log('Table does not exist, no need to delete')
+      return Promise.resolve();
```
jonasfj: Agian with the `Promise.resolve()`

bin/provisioner.js
```diff
@@ -1 +1,69 @@
-//TODO: The same as server but without configuring api...
```
jonasfj: If you look at the scripts I have else where they actually use `module.parent` to detected if they are top-level and only run `lunch` in that case... otherwise they just export the `launch` function.

This makes it possible to run it within the same process in the tests.

bin/upload-config-to-heroku.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+/* Take a taskcluster.base.config config file and generate a command that will
+   upload all values to heroku */
```
jonasfj: Do we really need this? Your local config should almost never the same as production.
And it's rarely that many variables to overwrite.. (Anyways, we still own it to rethink config system)

config/defaults.js
```diff
@@ -0,0 +1,42 @@
+module.exports = {
+  taskcluster: {
+    queueBaseUrl: 'https://queue.taskcluster.net/v1',
+    authBaseUrl: 'https://auth.taskcluster.net/v1',
```
jonasfj: Hmm... is there a good reason not to use the defaults from the taskcluster-client package. And just define them as `undefined` here.

config/defaults.js
```diff
@@ -0,0 +1,42 @@
...
+  server: {
+    publicUrl:  'https://localhost',
+    port:       5556,
+    env:        'development',
+    forceSSL:   false,
+    trustProxy: false,
+  },
+  azure: {
+    accountName: undefined,
+    accountKey:  undefined,
```
jonasfj: You don't need this anymore... You need it for your manage tables utility... but you really should use the API end-points on auth for getting temporary azure credentials for a specific table..

Take a look at how this done in the index:
https://github.com/taskcluster/taskcluster-index/blob/master/bin/server.js#L37-L42
See `auth.azureTableSAS` for details on scopes: http://docs.taskcluster.net/auth/api-docs/#azureTableSAS

config/defaults.js
```diff
@@ -0,0 +1,42 @@
...
+  },
+  aws: {
+    accessKeyId: undefined,
+    secretAccessKey: undefined
```
jonasfj: You might want to consider specifying API version... I don't know how important it is. We just do everywhere else to lock things down (bumping the API version is fairly trivial after all).

provisioner/aws-pricing.js
```diff
@@ -0,0 +1,93 @@
...
+var lodash = require('lodash');
+var util = require('util');
+var Cache = require('../cache');
+var assert = require('assert');
+
+
+function fetchPricing(ec2) {
+  var that = this;
+  var startDate = new Date();
+  startDate.setHours(startDate.getHours() - 2);
```
jonasfj: Hmm... Couldn't we cache something like 3 hours of data-points in memory and then just fetch from the last data-point we have...

provisioner/aws-pricing.js
```diff
@@ -0,0 +1,93 @@
...
+  this.__pricing = pricing;
+  this.regions = regions;
+}
+
+
+/**
+ * Get the pricing dictionary... This is a little ugly
+ * but it works... Ideas welcomed
+ */
+AwsPricing.prototype.get = function () {
```
jonasfj: `priceHistoryByRegion` if it's naming you're looking for...

You should probably return a clone. If you maintain state... I doubt performance is an issue here, granted if you have a tonn of calls to it, you might want to do something else. (Nothing, or try with `object.freeze`)

provisioner/aws-state.js
```diff
@@ -0,0 +1,292 @@
...
+        Values: [keyPrefix + '*']
+      }, {
+        Name: 'state',
+        Values: ['open']
+      }]
+    }),
+  ]);
+
+  p = p.then(function(res) {
+    return new AwsState(keyPrefix, _classify(ec2.regions, keyPrefix, res[0], res[1])
```
jonasfj: I know it's legal in javascript to reference a function prior to it's definition but it scares me shitless...
This is another reason I create objects. To very much avoid this.. where you too easily and up referencing undefined functions... This is actually the reason why I write everything function as `var myFunc = function() {.... };` so it can't be referenced prior to definition.

provisioner/aws-state.js
```diff
@@ -0,0 +1,292 @@
...
+var assert = require('assert');
+var debug = require('debug')('aws-provisioner:aws-state');
+
+
+/**
+ * Return a promise that resolves to an AWS State object.  We do this
+ * so that we can grab the state using promises but then inspect state
+ * using synchronus methods, as they're all just list processing
+ */
+function fetchState(ec2, keyPrefix) {
```
jonasfj: Why not just export the constructor for the `AwsState` object, and the put an `update` method on the `AwsState` object.

Then you validate config input in the constructor, set things like `ec2` and `keyPrefix` as properties on `this` and share them between methods that needs them.

provisioner/aws-state.js
```diff
@@ -0,0 +1,292 @@
...
+  workerType.listRegions().forEach(function(region) {
+    var rState = that.__state[region]; 
+
+    if (!rState[wName]) {
+      return;
+    }
+    
+    var wState = rState[wName];
+
+    if (states.indexOf('running') !== -1) {
```
jonasfj: How about:
```
states.forEach(function(state) {
  wState[state].forEach(function(instance) {
    capacity += workerType.capacityOfType(instance.InstanceType);
  });
});
```

provisioner/aws-state.js
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Count the capacity of this workerType that are in the states specified
+ * by `states`.  Doing this uses the Capcity key from the workerType's
+ * types dictionary.  Remember that capacity is the number of tasks
+ * that this instance/request will be able to service.
+ * If specified, `extraSpotRequests` is a dictionary which contains a region
+ * and worker type categorized list of outstanding spot requests
+ */
+AwsState.prototype.capacityForType = function(workerType, extraSpotRequests, states) {
```
jonasfj: Again, a good note, if you tracked all aws state in `AwsState` and used an `update` method, you wouldn't have to inject `extraSpotRequests`. And the data structure for this state object would be strictly limited the `AwsState` object. Granted one would have to notify the `AwsState` object when submitting a spot request. Or do the spot request using a method on `AwsState` like `AwsState.requestSpotInstance(launchConfiguration)`. Then create the launchConfig, etc. somewhere else..

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
+'use strict';
 var base        = require('taskcluster-base');
 var assert      = require('assert');
 var Promise     = require('promise');
-var _           = require('lodash');
+var lodash      = require('lodash');
```
jonasfj: Why, not `_` ?

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+     * {
+     *   'us-west-1': {
+     *     'overwrites': {
+     *       'ImageId': 'ami-aaaaaaa'
+     *      }
+     *   }
+     * } */
+    regions: base.Entity.types.JSON,
+  },
+  context: ['ec2', 'keyPrefix', 'pubKey', 'influx'],
```
jonasfj: You problably don't need `influx` it might be useful with a `reporter` see `base.stats` on how to create a reporter when you have a stats drain like an instance of `Influx`.

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+
+
+/**
+ * Return an Object for JSON encoding which represents
+ * the data associated with this WorkerType.  This is a
+ * method intended for use in displaying the data associated
+ * with a given workerType
+ */
+WorkerType.prototype.json = function() {
+  return lodash.clone(this.__properties);
```
jonasfj: Comments on accessing `__properties` from last review still stands. It's much better to be explicit.

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+};
+
+
+/**
+ * Load a single workerType by name
+ */
+WorkerType.load = function(workerType) {
+  assert(workerType);
+  return base.Entity.load.call(this, {
+    workerType: workerType
```
jonasfj: Again, can we not just use the `base:Entity.load` syntax where we provide a dictionary. I know it not as fun, but it's little overhead and gives us a consistent API.

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+    Filters: [{
+      Name: 'key-name',
+      Values: [keyName]
+    }] 
+  });
+
+  p = p.then(function(res) {
+    var toDelete = [];
+
+    that.listRegions().forEach(function(region) {
```
jonasfj: Please do a `return Promise.all(that.listRegions().map(...`

I don't care if you split it... but using `forEach` and `push` isn't much fun... `filter` + `map` fits great with `Promise.all`.

In this case you might do:
```
var toDelete = that.listRegions().filter(function(region) {
  return res[region].KeyPairs[0];
});
return Promise.all(toDelete.map(function(region) {
  return that.ec2.deleteKeyPair.inRegion(region, {KeyName: keyName});
});
```

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+}
+
+
+/**
+ * We need to be able to create a launch specification for testing without
+ * already having an instance.  This is available as a non-instance method
+ * so that we can create and test launch specifications before inserting
+ * them into Azure.
+ */
+WorkerType.createLaunchSpec = function(debug, region, instanceType, worker, keyPrefix) {
```
jonasfj: This is horribly ugly, but I love it... It looks to produce some really good and fairly readable error messages :)
(besides there is no way this can be pretty).

Depending on how or where you use this might want to returns errors as JSON instead.. At least that is more useful when API handers, but I suspect that is what the test one is for.

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+
+/**
+ * We use KeyPair names to determine ownership and workerType
+ * in the EC2 world because we can't tag SpotRequests until they've
+ * mutated into Instances.  This sucks and all, but hey, what else
+ * can we do?  This method checks which regions have the required
+ * KeyPair already and creates the KeyPair in regions which do not
+ * already have it
+ */
+WorkerType.prototype.createKeyPair = function() {
```
jonasfj: I would be tempted to move key management into `AwsState´ or do it in separate object. Mostly because it seems like it makes sense to refresh it every 20 min (perhaps) and just track state in-memory.

And key creation/deletion logic seems out of place here..
That say it might make sense to a add a utility function like `WorkerType.prototype.region()` that that returns a list of regions the workerType exists in... Abstracting away the `Object.keys()` and potential filtering.

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+
+  return p;
+
+};
+
+/**
+ * Shutdown all instances of this workerType, cancel
+ * any open spot requests.
+ */
+WorkerType.prototype.killAll = function(debug) {
```
jonasfj: I'm not sure when you use this... But it seems tempting to me to do it in `AwsState`..
If it's for the case where you delete the workerType, it seems to me like it's not a special case.

I might not have made this clear. But I consider the aws-provisioner to own all ec2 instances that has a keyName starting with `aws-provisioner-managed-` (or whatever prefix the provisioner is configured with). And we should terminate spot requests and instances with this key, but an invalid `workerType` ie. a workerType that doesn't exist because it was deleted etc...

provisioner/data.js
```diff
@@ -1,30 +1,704 @@
...
+  return spotBids;    
+};
+
+
+/**
+ * Create an instance of this workerType given a bid.
+ * A bid is an object with the keys of `region`, `price`
+ * and `type`
+ */
+WorkerType.prototype.spawn = function(debug, bid) {
```
jonasfj: Again, I would be tempted to put this in AWS state.

provisioner/provision.js
```diff
@@ -0,0 +1,302 @@
...
+  function provisionIteration() {
+    // We should cancel the last iteration's watch dog
+    if (that.__watchdog) {
+      clearTimeout(that.__watchdog);
+    }
+    // And make sure we set this one!
+    that.__watchdog = setTimeout(function() {
+      generalDebug('KILLING PROVISIONER BECAUSE IT APPEARS TO BE STUCK...');
+      // Hmm, should I instead just process.exit(1);
+      throw new Error('PROVISIONER HAS FALLEN AND CAN\'T GET BACK UP');
```
jonasfj: This is pretty neat... I usually include the string `[alert-operator]` in debug messages when I want to be notified...Then I configure an alart on papertail.

Anyways, this is awesome! :)

provisioner/provision.js
```diff
@@ -0,0 +1,302 @@
...
+
+
+
+/**
+ * Provision a specific workerType.  This promise will have a value of true if
+ * everything worked.  Another option is resolving to the name of the worker to
+ * make it easier to see which failed, but I'd prefer that to be tracked in the
+ * caller. Note that awsState as passed in should be specific to a workerType
+ */
+Provisioner.prototype.provisionType = function(debug, workerType, state, pricing) {
```
jonasfj: Note, if `AwsState` just tracked the state and had an `update` mehtod as previously suggested, you could have it as a property on `this`. Same thing with `pricing`...

provisioner/provision.js
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Because the AWS is eventually consistent, it will sometimes take time for
+ * spot requests to show up in the describeSpotInstanceRequests calls for
+ * AWS state.  We will maintain an internal table of these submitted but
+ * not yet visible spot requests so that we can offset the count of a given
+ * instance type for figuring out running capacity.  If the provisioning
+ * process is restarted before the spot request shows up in the api's
+ * state we will lose track of it until it turns into an instance.
+ */
+Provisioner.prototype.addNewSpotRequests = function(debug, state, newSrs) {
```
jonasfj: Again, keep state in `AwsState` and putting this into `AwsState` would mean that a lot of this code could be isolated. Rather than us maintaining and passing complex data structures around.

jonasfj:
Pretty good, so far... Structure is not all bad. You have functions that returns idempotent objects. Like `AwsState`, I think you would be better off mutating those objects with an `update` function. Even if you don't reuse state from previous request and just throw all of that away when `update` is called.

It'll be much better when we do more advanced state tracking in the future. For example what you're already doing with spotRequestIds which would be best served by being abstracted away be `AwsState` and the logic and data structures tracking this not exposed.

jonasfj:
Oh, now I remember why I wanted to write a comment :)

This is actually important. Maybe I'm missing something but where do kill renegade instances?
If an instance has keyName `aws-provisioner-managed-...` and it's in the wrong region, it's for a workerType that doesn't exists or it's over the capacity of the workerType it must be killed!
Even if it's over the spotBid for the workerType or has wrong AMI, not sure how far we should go with this.

But I'm sure that if instances shows up for a workerType we don't have, we should kill it! Note, we track instances owned by aws-provisioner using keyName prefix.

routes/v1.js
```diff
@@ -7,38 +7,85 @@ var base        = require('taskcluster-base');
...
+    "type.  Based on the number of tasks pending and the scaling ratio, we may",
+    "submit requests for new instances.  We use pricing information, capacity and",
+    "utility factor information to decide which instance type in which region would",
+    "be the optimal configuration.",
+    "",
+    "Each EC2 instance type will declare a capacity and utility factor.  Capacity is",
+    "the number of tasks that a given machine is capable of running concurrently.",
+    "Utility factor is a relative measure of performance between two instance types.",
+    "We multiply the utility factor by the spot price to compare instance types and",
+    "regions when making the bidding choices.",
```
jonasfj: +1 for writing docs, I love it... I know it's hard to write this stuff... But it's very useful for have :)

jonasfj:
Okay, I'll admit to have worn out when I came to api... I looks somewhat like what it did last time..
If you want I'll take a better look at that later...

Most notably, I think this is starting to look good. The overall structure is fairly solid. There probably a bit too much logic in `WorkerType` I would move the aws specific things to `AwsState` and make that object mutable. Some of the price decision logic and provision logic could also be moved to `provisioner` from `WorkerType`. But I love the validation methods in `WorkerType` and the `createLaunchSpec` makes perfect sense there. The `capacityOfType` is really great, isolates data access and makes it simpler.
(Not that isolating data access is important, but for something it's neat).

jonasfj:
Final note, do run this under `babel` (formerly `6to5`) setting it up is very simple. You can checkout the queue rewrite, it's basically just the mocha hook and the using `babel-node -r` instead of `node` in scripts section of `package.json`.

Your code is so perfectly well structured that introducing `await` would make it even prettier. And you don't even need to port all your code in one go. You can just run under `babel-node -r` and then slowly start using the features.

It's by far the best way to write functions that returns promises...

TODO.md
jhford: The hello endpoint is an api endpoint that currently doesn't do anything but is intended to be passed through UserData to the Workers.  It's what we discussed so that we could get metrics on how long things take to go from "please create an instance" to "instance ready".  The InstanceId scope, which is different to the WorkerId scope, would be used to ensure that only a specific instance can say hello for itself.

TODO.md
jhford: Yep, I'm writing that today but it's not going to change any major architectural things.

jhford:
When I originally split the state functions into their own object, I did have it where you'd create the state object once then there was an update function which returned a promise which resolved when the state was updated.  It worked but I encountered a bunch of really strange issues, which is why I moved to returning an object which was purely for inspecting the state.

I'd agree that `.spawn()` really ought not to be in the WorkerType, and the KeyPairs functions aren't totally in the right place there.  If we move those to the AwsState object, we're really expanding the scope of that object well beyond the state of the system.  If we do move these methods there, I think I'll rename AwsState to something more general.  I like this idea.  My main reason for moving away from this model was that I didn't want the risk of having two different provisioning iterations accessing the state at the same time and one getting out of date information.  Now that we have the `.run()` method and we aren't calling the `provisionAllTypes` method outside of the Provisioner class, I'm a little less scared.

Fetching and storing more pricing data is definitely something we could do, but unless we're storing more than a day's worth of storage history to get long term trends, I think we're better off looking at a shorter period of time and avoiding the added complexity of maintaining that state.  I also think that is a complex enough thing that I'd want to split aws pricing things into their own component.

jhford:
afd21d38797c4ac7643b5c59c4c8b2974af80ea9 and 9ee1b36ba9c203d1b0a0b358902cf129a22c70b0 rewrite the management script, which:

* switches to commander from the jankfest that was happening previously
* adds a delete-all which means we don't really need to be able to delete tables
* gets rid of the manageTable.js script
* gets rid of the older scripts which we don't need anymore
* moves creating the table into the manage.js script
* switch to using auth.tc.net's table assertion endpoint instead of raw credentials

bin/manage-worker-types.js
jhford: done.  I've rewritten this utility with commander and it's much nicer now

bin/manageTable.js
jhford: This script is in the big bit bucket in the sky... This was left over from well before we talked about this, I just never really looked at the script so long as it worked.

bin/upload-config-to-heroku.js
jhford: It's helpful for me as I do development locally.  I think it's a generally useful script because it lets you mirror your local environment on a heroku instance easily.  If you're against it, I can remove it from here.

config/defaults.js
jhford: I guess not.  They were defined in one of the packages that I used to refer to for setting up the configuration.

provisioner/data.js
jhford: i just don't like that name, i think it makes things ugly.

provisioner/data.js
jhford: I agree that it's better to be explicit, but as things are right now, we'd need to define and maintain the same information in at least three different places:

* WorkerType.configure call
* this function
* json schema

This seems like a general thing we'd want in many places.  Why not add a `.asJson()` method to base.Entity.  We could even do fun things like ensuring that the thing that's returned matches the keys and types as defined in `.configure()`.

provisioner/data.js
jhford: I've switched to this.  I don't really see it being any better but if it makes you happy, it's no big deal to me

provisioner/data.js
jhford: Fixed

provisioner/data.js
jhford: This is moved to the new `AwsManager` class.  This class tracks in memory which keys it knows are present.  It will only do the API describe keys call if the key is not found in the memory cache.

Refreshing this every 20m would be a bad idea... It means that we could theoretically fail the first 20 minutes of provision of a new workerType because the key hasn't been created.

provisioner/data.js
jhford: Moved to `AwsManager`.  Implemented a rouge killer.  Implemented the kill everything logic using the new rouge killer logic.

provisioner/data.js
jhford: Done.

provisioner/provision.js
jhford: I can add that string, but I'd prefer not to.  Relying on magic strings in debugging message is a recipe for disaster as has been demonstrated to me through experience.  I think the solution is to move away from the debug module and have real logging in our applications... I think the provisioner at least is complex enough to benefit from real logging.

provisioner/provision.js
jhford: For now, I'd like to leave the pricing as it is.  I moved creating the `AwsManager` into the bin/provisioner script and am now injecting it into the `Provisioner`.  This makes it way easier to test the Provisioning logic should we want to.

provisioner/data.js
jonasfj: `base.Entity` does support non-json types like `Date` objects and buffers.

Besides is very rare that what we store in table storage matches what we return... Ususally we do some modifications because we store in a slightly different format, or we don't store redundant information...In almost all cases we have internal state that we don't return the users.

I suspect it's just a matter of time before you'll have that here too.

provisioner/provision.js
jonasfj: Yeah.. real logging is one of the things that we would like to add... Maybe we can cook a setup into taskcluster-base in a way that can be reused in other components. Perhaps it's one of those cases were we really wan to use global scope, I'm not sure. So far I've been postponing real logging to next time we rework our platform using something like `hapi`.

PR #8, opened 2015-03-17
I think we're ready for deployment!

jhford:
Hey, can you take another look at the provisioner?  I have it running on heroku and it's going pretty well there.

Please let me know what you think


watchdog.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
```
jonasfj: How about we put this file somewhere else... perhaps `provisioner/` I typically put all logic into folders like that... `lib/` is another option, but I fear the file would be lonely in there...Or it share it with `cache.js`.

watchdog.js
```diff
@@ -0,0 +1,50 @@
...
+ * `WatchDog` will throw an `Error` with `msg` if the
+ * timer is allowed to expire
+ */
+function WatchDog(maxTime, action) {
+  this.maxTime = maxTime;
+  if (typeof action === 'function') {
+    this.action = action;
+  } else {
+    this.action = function() {
+      throw new Error(action);
```
jonasfj: It might be preferable to emit and `error` event on the object instead... if not cached it'll cause node to crash...

package.json
```diff
@@ -1,29 +1,35 @@
...
   "devDependencies": {
-    "mocha":                            "2.0.1",
-    "slugid":                           "1.0.3"
+    "mocha": "2.0.1",
+    "should": "^4.3.0",
+    "slugid": "1.0.3"
   },
   "engines": {
-    "node":                             "0.10.x"
+    "node": "0.10.x"
```
jonasfj: Let's move to node 0.12, unless there is a good reason not to.

And let's lock down dependencies with `npm shrinkwrap` (this should be done before deployment).

package.json
```diff
@@ -1,29 +1,35 @@
 {
-  "name":             "taskcluster-aws-provisioner",
-  "version":          "0.1.0",
-  "private":          true,
-  "main":             "bin/server.js",
+  "name": "taskcluster-aws-provisioner",
+  "version": "0.1.0",
+  "private": true,
+  "main": "bin/server.js",
```
jonasfj: You'll probably need `babel-node -r` here.

test/mocha.opts
```diff
@@ -1,3 +1,3 @@
+--require should
 --ui tdd
 --timeout 30s
```
jonasfj: You should definitely add `--require test/babel-register.js`

And add a file like this:
https://github.com/taskcluster/taskcluster-queue/blob/rewrite-on-azure-storage-services/test/6to5-register.js

So that everything will be compiled on the fly....Otherwise tests won't work when you start using babel.

routes/v1.js
```diff
@@ -7,38 +7,85 @@ var base        = require('taskcluster-base');
...
+    "submit requests for new instances.  We use pricing information, capacity and",
+    "utility factor information to decide which instance type in which region would",
+    "be the optimal configuration.",
+    "",
+    "Each EC2 instance type will declare a capacity and utility factor.  Capacity is",
+    "the number of tasks that a given machine is capable of running concurrently.",
+    "Utility factor is a relative measure of performance between two instance types.",
+    "We multiply the utility factor by the spot price to compare instance types and",
+    "regions when making the bidding choices.",
+    "",
```
jonasfj: nit, line length exceeded 80 chars.

routes/v1.js
```diff
@@ -7,38 +7,85 @@ var base        = require('taskcluster-base');
...
+
+function errorHandler(err, res, workerType) {
+  console.error(err, err.stack);
+  switch(err.code) {
+    case 'ResourceNotFound':
+      return res.status(404).json({
+        message: workerType + ': ' + err.body.message.value,
+        error: {
+          workerType: workerType,
+          reason: err.code,
```
jonasfj: It's a bad idea to expose error codes from azure, fact that we azure table storage is internal, consumers have no business knowing that.
Message should be human readable and explain `"workerType: " workerType + " wasn't found",

Note, we should formalize errors, there is a bug on the issue, feel free to tune in:
https://bugzilla.mozilla.org/show_bug.cgi?id=1141639

routes/v1.js
```diff
@@ -46,99 +93,131 @@ api.declare({
...
-    return; // by default req.satisfies() sends a response, so we're done
+    return; // by default req.satisfies() sends a response on failure, so we're done
   }
 
-  // TODO: If workerType definition specifies scopes that should be given
+  // TODO: If workerType launchSpecification specifies scopes that should be given
   //       to the workers using temporary credentials, then you should validate
   //       that the caller has this scopes to avoid scope elevation.
+  // TODO: ^ do this.  not entirely sure what this means, are LaunchSpecification
+  //       security groups the same as scopes?
```
jonasfj: Nope....

We should add a list of scopes that should given in temporary credentials to workers created with the launchSpecification. When we do this, we should require this scopes from the person creating the workerType to avoid scope elevation issues.

jonasfj: Obviously you're not issuing temporary credentials yet...

schemas/get-worker-type-response.yml
```diff
@@ -0,0 +1,112 @@
...
+#        overwrites:
+#          type: object
+#          description: |
+#            This object contains keys which will overwrite the default AWS Launch
+#            Specification with values that are unique to this instance type.
+#      required:
+#        - capacity
+#        - utility
+#        - overwrites
+#      additionalProperties: false
```
jonasfj: This is easily done with:
```
types:
  type: object
  description: ...
  additionalProperties:
    type: object
    properties:
       capacity: ....
```

jonasfj: Note, `types` should be `instanceTypes` and let's make it an array. It's a lot easier to read and understand the documentation... And easier to code against... I'm sure you'll eventually want it internally too. But for now you can just convert it...

Arrays are better validated by JSON schema too... as you can make the `instanceType` key an `enum`.

schemas/get-worker-type-response.yml
```diff
@@ -0,0 +1,112 @@
...
+#          This object contains keys which will overwrite keys in the default and
+#          instance-type specialized Launch Specifications.
+#        properties:
+#          ImageId:
+#            type: string
+#            description: |
+#              Because AMI IDs (ImageId in the API Objects) are unique to each region
+#              we force this key to be present
+#        required:
+#          - ImageId
```
jonasfj: Why is this commented out, and not using `additionalProperties: false`.

As with `types` I would make this an array... Even if just for external tools for now...

schemas/get-worker-type-response.yml
```diff
@@ -0,0 +1,112 @@
+id:       http://schemas.taskcluster.net/aws-provisioner/v1/get-worker-type-response.json#
```
jonasfj: Instead of calling this file: `get-worker-type-response.json` how about `worker-type.json`.

The create/get requests can both use the same schema. And if you need to embed in another schema you can use `$ref` construct... Or the special `contants.js` file... There is no need for duplication of JSON schemas.. We have plenty of tools make it generic.

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+
+// Delete workerType
+// TODO: send a pulse message that a worker type was removed
+api.declare({
+  method:         'delete',
+  route:          '/worker-type/:workerType',
+  name:           'removeWorkerType',
+  deferAuth:      true,
+  // TODO: Should we have a special scope for workertype removal?
+  scopes:         ['aws-provisioner:get-worker-type:<workerType>'],
```
jonasfj: Let's just require `aws-provisioner:remove-worker-type:<workerType>`

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+  return p;
+});
+
+
 api.declare({
   method:         'get',
   route:          '/list-worker-types',
   name:           'listWorkerTypes',
-  deferAuth:      true,
+  deferAuth:      true, // I don't think we need this unless we do parameterized scopes
```
jonasfj: Correct...

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
-    // Document what this method does
+    "List all known WorkerType names",
+  ].join('\n')
+}, function(req, res) {
+  var ctx         = this;
+  var workerType  = req.params.workerType;
+
+  if(!req.satisfies({
+    workerType:       workerType
+  })) {
```
jonasfj: Set `deferAuth: false` (or just don't define `deferAuth` above, it defaults to `false`), then this check is no longer necessary.

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
   input:          undefined,  // No input
   output:         SCHEMA_PREFIX_CONST + 'list-worker-types-response.json#',
   title:          "List Worker Types",
   description: [
-    // Document what this method does
+    "List all known WorkerType names",
+  ].join('\n')
+}, function(req, res) {
+  var ctx         = this;
+  var workerType  = req.params.workerType;
```
jonasfj: No such URL parameter is given...

routes/v1.js
```diff
@@ -46,99 +93,131 @@ api.declare({
...
-    // Document what this method does
+    "Create a worker type and ensure that all EC2 regions have the required ",
+    "KeyPair"
   ].join('\n')
-}, function(req, res) {
+}, 
+function(req, res) {
   var ctx         = this;
   var input       = req.body;
   var workerType  = req.params.workerType;
```
jonasfj: I know this sucks... and in future if we migrate to `hapi` this can be done at a higher level.
But I advice that we validate that `workerType` satisfies `/^[a-zA-Z0-9-_]{1,22}$/` at least when creating worker types...

schemas/list-worker-types-response.yml
```diff
@@ -0,0 +1,7 @@
+---
+  id: "http://schemas.taskcluster.net/aws-provisioner/v1/list-worker-types-response.json#"
+  title: "List Worker Types"
+  type: "array"
+  items: 
+    type: "string"
```
jonasfj: `pattern: '^[a-zA-Z0-9-_]{1,22}$'` to ensure that they are valid identifiers.

In taskcluster we use slugid which matches `/^[a-zA-Z0-9-_]{22}$/` and human readable identifiers for a lot of other things, these matches `/^[a-zA-Z0-9-_]{1,22}$/`. Small is powerful when we need to stick into SSH keynames (on EC2), azure queue names, rabbitmq routingkeys, urls, and so forth...

The choice may seem a bit arbitrary, but it's a fairly well support character set... It's small, and can be decoded as URL-safe base64, if one wants to encode it efficiently in another format, for example base32.

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+
+  return p;
+
+});
+
+
+api.declare({
+  method:         'get',
+  route:          '/get-launch-specs/:workerType',
+  name:           'getLaunchSpecs',
```
jonasfj: Am I the only one tempted to call this `testLaunchSpecs`...

schemas/get-aws-state-response.yml
```diff
@@ -0,0 +1,6 @@
+id:       http://schemas.taskcluster.net/aws-provisioner/v1/get-aws-state-response.json#
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Get AWS State Response"
+description: |
+  All of the launch specifications for a worker type
+type:                       object
```
jonasfj: Should this be array of objects?
IMO, we don't have to declare what properties the objects have, if we just say that they are in a format compatible with the EC2 format from aws-sdk. 

Note, this is a low priority end-point. So another alternative is to just add in the docs a note that says `**This API end-point is experimental and may be subject to change without warning.**`.
Then we don't have to feel bad about changing it..

Note, if you feel like it, I filed bug 1144261 for adding an `stability: 'experimental'` option to `api.declare` this could be useful in many cases. We could probably use node.js stability levels (granted I'm not sure I ever want to promise stable :)
Please share thoughts: https://bugzilla.mozilla.org/show_bug.cgi?id=1144261

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+  scopes:   [[
+    'aws-provisioner:aws',
+    'aws-provisioner:get-worker-type:<workerType>',
+  ]],
+  description: [
+    "WARNING: YOU ALMOST CERTAINLY DO NOT WANT TO USE THIS ",
+    "Shut down every single EC2 instance associated with this workerType. ",
+    "This means every single last one.  You probably don't want to use ",
+    "this method, which is why it has an obnoxious name.  Don't even try ",
+    "to claim you didn't know what this method does!"
```
jonasfj: Can we add notice that this is experimental.
```js
    "",
    "**Warning, this API end-point is experimental and may change without warning.**"
```

I dont' like the `route` or the `name`, but appreciate the length/complexity/clarity of the `name`.
Perhaps `route: '/worker-type/:workerType/terminate-all-instances'` would be more in style with what we do elsewhere.
For name maybe `terminateAllInstancesOfWorkerType`... The "i" and "L" following each other this way will certainly make it hard to type :)

IMO, it's more important to give this an unreasonable scope...

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+
+
+api.declare({
+  method:   'get',
+  route:    '/shutdown/every/ec2/instances/for/:workerType',
+  name:     'shutdownEveryEc2InstanceFor',
+  title:    "Shutdown Every Ec2 Instance of this Worker Type",
+  scopes:   [[
+    'aws-provisioner:aws',
+    'aws-provisioner:get-worker-type:<workerType>',
```
jonasfj: Instead of `aws-provisioner:get-worker-type:<workerType>` how about `aws-provisioner:terminate-all-instances-worker-type:<workerType>`

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+    errorHandler(err, res, workerType);
+  });
+
+  return p;
+
+});
+
+
+api.declare({
+  method:   'get',
```
jonasfj: At the very least, let's make this a `POST` request :)

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+    });
+  });
+
+  return p;
+
+});
+
+
+api.declare({
+  method:   'get', // Hmm, maybe this should be post
```
jonasfj: ha ha, yes maybe... :)

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+  deferAuth:      true,
+  scopes:         [[
+    'aws-provisioner:get-worker-type:<workerType>'
+  ]],
+  output:         SCHEMA_PREFIX_CONST + 'worker-type-stats-response.json#',
+  title:          "Spawn an instance of a Worker Type",
+  description: [
+    'Spawn an instance of a workerType given a bid.  A bid is an ',
+    'object which has a number `price`, string `region` and string ',
+    '`instance type`.'
```
jonasfj: Update docs...

jonasfj: declare experimental if you're not defining a more clear JSON schema... We probably don't critically need this now.

routes/v1.js
```diff
@@ -149,37 +228,483 @@ api.declare({
...
+});
+
+
+api.declare({
+  method:         'get',
+  route:          '/worker-type/:workerType/stats',
+  name:           'workerTypeStats',
+  deferAuth:      true,
+  scopes:         [[
+    'aws-provisioner:get-worker-type:<workerType>'
```
jonasfj: Okay, this end-point is not important at this point... And it's hard to see what it returns...

Anyways, if only returns numbers for each workertypes, they maybe we can make it public...

jonasfj:
Okay, I finished with `v1.js` you can definitely start addressing those comments.
I assume most of the API end-points should be declared experimental.
@jhford, can we agree on that.

I'm fear that garndt wants to use the API in his worker CI stuff as soon as it's out. So it's best if it's clear which API end-points aren't going to change. In particular the kill all and API end-points exposing internals should be declared experimental.

Am I right in saying that the only relatively stable end-points are create/update/remove workerTypes.
(After we move arrays, rename `types` to `instanceTypes` and define propers on objects for instances types and regions).

watchdog.js
jhford: Done.  Moved `cache.js` and `watchdog.js` into a new `lib/` directory.  In general, I feel like most of the things in the `provisioner/` directory really belong here.  It's confusing that each project's lib directory is not called lib.

watchdog.js
jhford: I've modified the WatchDog to be an eventemitter.  In `provisioner/provision.js`, though, I'm throwing an error still on the `expired` event.  Crashing node in this case is the desired behaviour, unless you know of a way that I can 'cancel' a promise from outside the promise...

package.json
jhford: Done

package.json
jhford: I... don't think so.  My understanding is that this is just the file that node ought to look for when requiring this whole package when installed as a dependency.

test/mocha.opts
jhford: done.  makes sense

routes/v1.js
jhford: I don't respect the 80 char limit anywhere else in the program really... I like verbose names, so 120 is what I set my editors to normally.

bin/provisioner.js
```diff
@@ -1 +1,95 @@
...
+      keyPrefix: keyPrefix,
+      provisionerId: provisionerId,
+    },
+  });
+
+  // Create all the things which need to be injected into the
+  // provisioner
+  var ec2 = new aws('EC2', cfg.get('aws'), allowedRegions);
+  var awsManager = new AwsManager(ec2, keyPrefix, pubKey);
+  var queue = new taskcluster.Queue({credentials: cfg.get('taskcluster:credentials')});
```
jonasfj: nit, line is wider than 80 chars

watchdog.js
jonasfj: yeah we should clean this up... and make a sane pattern.
When we do the next cleanup iteration of all component, move to conf, glue code, hapi and other things we have pending...

package.json
jonasfj: yeah, that's right.. so maybe we just shouldn't even have this :)

routes/v1.js
jhford: fixed.  I agree that we should standardise error messages, but the provisioner is not the place to do it :)

routes/v1.js
jhford: wow, that comment is so out of date.... I don't know why it's still there...

routes/v1.js
jonasfj: I don't really care about 80 vs. 120... but I strongly prefer consistency and we have 80 chars everywhere else.
It's also the general style guideline as I see it:
https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Line_Length
I see no need to rush to fix it... but going forward I think 80 is the sane standard we do everywhere.
If you want to define a style guide I'm open to that, but I like being as close to the mozilla coding style guideline as possible; granted it's not very specific about javascript.

I just noticed that we recommend javadoc tags... So I guess that settles my documentation issues. If I can find something that can document ES6/7 with that kind of tags. Otherwise I'll probably keep using the obscure/undefined strategy we have now :)

routes/v1.js
jonasfj: Nope, the standardization should happen in `base.API`. However, please tune in on the bug with any considerations you have. Personally I would prefer a very small list of error codes that we can reuse between components:
https://bugzilla.mozilla.org/show_bug.cgi?id=1141639

routes/v1.js
jhford: It's weird that we'd have a different set of scopes for creating, updating, viewing and removing.

Shouldn't we only have two: view and mutate?  Why would we ever let someone who can create a workertype not update it, or let someone remove one which they cannot create....

routes/v1.js
jhford: removed this and the call to satisifes()

routes/v1.js
jhford: done

routes/v1.js
jhford: removed

routes/v1.js
jhford: Done.  I am now asserting that in the WorkerType.create method.  Errors there will bubble up to the response.

routes/v1.js
jhford: Well, you could call it that, but it both tests and returns all the launch specs.  since it's both a tester and a getter, I think it's more logical for a getter to be used to test than a tester to get.

routes/v1.js
jonasfj: hmm... that's  good point... I see a potential use-case for some automation process that can updated, but not create/remove worker types.

But we can always add another scope set and maintain backwards compatibility, if we want to, so let's just use:
 * `aws-provisioner:manage-worker-type:<workerType>` (view/modiy)
 * `aws-provisioner:get-worker-type:<workerType>` (view-only)
(I'm also open to better names) 

For the get workertype operation the scopeset would then be:
```js
scopes: [
  ['aws-provisioner:get-worker-type:<workerType>'],
  ['aws-provisioner:manage-worker-type:<workerType>']
]
```

For the create/update/remove it would be:
```js
scopes: [
  ['aws-provisioner:manage-worker-type:<workerType>']
]
```

Note, this is stuff we can change relatively easy... without breaking compatibility.

schemas/get-aws-state-response.yml
jhford: I've removed a couple of endpoints which I don't really have a strong reason for continuing to include for now.  I think having a stability keyword is cool, as long as it includes 'deprecated'.  It'd be cool if we automatically marked those as deprecated in the client using https://nodejs.org/api/util.html#util_util_deprecate_function_string

jonasfj: Added to bug, that would be totally cool to use: https://nodejs.org/api/util.html#util_util_deprecate_function_string

routes/v1.js
jhford: Ok, I've renamed it as you suggested. I've also changed it to requiring the same scopes as the turn absolutely everything off endpoint, since really, no one but super users should be doing this.

routes/v1.js
jhford: Instead, let's just make it so only the super user can do it.

routes/v1.js
jhford: or just delete it altogether... We can add it back simply enough if we really want it.

config/production.js
```diff
@@ -1,3 +1,18 @@
...
+    awsKeyPrefix:         'aws-provisioner2-managed:',
+    iterationInterval:    1000 * 75,
+    allowedRegions:       'us-west-1,us-west-2,us-east-1', // comma seperated list
+  },
+  server: {
+    publicUrl:  'https://aws-provisioner2.herokuapp.com',
+    port:       5556,
+    env:        'production',
+    forceSSL:   false,
+    trustProxy: false,
```
jonasfj: forceSSL = true, same with trustProxy. as we run behind a trusted reverse proxy on heroku.

routes/v1.js
jhford: The original idea was to expose the list of InstanceIds and SpotInstanceRequestIds from the AWS api.  In the end, I'm probably going to do this slightly differently.

config/production.js
```diff
@@ -1,3 +1,18 @@
 module.exports = {
-
-};
\ No newline at end of file
+  provisioner: {
+    id:                   'aws-provisioner2',
+    workerTypeTableName:  'AWSWorkerTypesV2',
+    publishMetaData:      'false',
+    statsComponent:       'aws-provisioner2',
+    awsKeyPrefix:         'aws-provisioner2-managed:',
```
jonasfj: the `2` has to go when we go into production...
Granted we might want to come up with a non-conflicting awsKeyPrefix... perhaps `provisioner-managed:`

jhford:
> Okay, I finished with v1.js you can definitely start addressing those comments.
I assume most of the API end-points should be declared experimental.
@jhford, can we agree on that.

Great.  Most of the ones that would be experimental have been removed... They just aren't worth the hassle right now.  I've marked the others with experimental.

> I'm fear that garndt wants to use the API in his worker CI stuff as soon as it's out. So it's best if it's clear which API end-points aren't going to change. In particular the kill all and API end-points exposing internals should be declared experimental.

> Am I right in saying that the only relatively stable end-points are create/update/remove workerTypes.
(After we move arrays, rename types to instanceTypes and define propers on objects for instances types and regions).

I think that create, remove, update, retrieve are pretty solid at this point.  The shutting down endpoint should be pretty solid, but I'm marking it as experimental nonetheless.

jhford:
Ok, I think I've addressed most issues aside from:

* renaming types -> instanceTypes
* exposing regions and instanceTypes as lists instead of objects

package.json
jhford: rgr, deleting

provisioner/aws-pricing.js
```diff
@@ -0,0 +1,146 @@
...
+    Array.prototype.push.apply(zones, that.availableAvailabilityZonesInRegion(region)); 
+  });
+  return zones;
+};
+
+
+/**
+ * List the available availability zones in one region
+ */
+AwsPricing.prototype.availableAvailabilityZonesInRegion = function(region) {
```
jonasfj: I strongly suspect that `availabilityZonesInRegion` is explicit enough... Why would we ever want to list anything not available...

provisioner/aws-pricing.js
```diff
@@ -0,0 +1,146 @@
...
+ * average to account for fluctuations in price.  A lot of times,
+ * the spot price will go from X to 7X then back to X.  Because we
+ * don't want to bid X, get killed when it goes to 7X, we should
+ * consider only the peak pricing.  Since we're only fetching a
+ * specified amount of time in the original request, we limit history
+ * that way.  Wouldn't it be nifty to have an arg to this function that
+ * let you specify (regardless of fetched history) the earliest 
+ * point in time to consider for the max finding
+ */
+AwsPricing.prototype.maxPrices = function() {
```
jonasfj: Generally I love the docs... they are pretty good many places...
Here an example of output would be nice... BUT..

I don't understand why you have `maxes[region][type][zone]`.
As far as understand aws-sdk, it returns availability zones they are always returned on the format `us-east-1a`.
Ie. globally unique.

I guess you might be doing this... because we actually want to filter them by regions allowed in workerType definition which doesn't specify the `-1a` part... Okay, maybe it makes sense... :)

/me reads on...

bin/provisioner.js
```diff
@@ -1 +1,95 @@
...
+    provisionerId: provisionerId,
+    taskcluster: cfg.get('taskcluster'),
+    influx: influx,
+    awsManager: awsManager,
+    pricingCache: pricingCache,
+    provisionIterationInterval: cfg.get('provisioner:iterationInterval'),
+  }
+
+  var provisioner = new provision.Provisioner(config);
+  provisioner.run();
```
jonasfj: nit, `return  provisioner.run()` would be neat.. So we can access it from tests :)

jonasfj:
@jhford,
I notice that you use `__` to denote private members. I usually do that if I expect someone to subclass and want to keep it sealed from subclasses... Otherwise I just use `_`...

Anyways, I don't care... but would live it if we could strive for consistency on the issue. No need to fixing anything though... Just saying :)

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+AwsManager.prototype.killCancel = function(region, instances, requests) {
+  assert(instances || requests);
+  var that = this;
+
+  var promises = [];
+  var i = instances || [];
+  var r = requests || [];
+
+  if (i.length > 0) {
+    debug('killing instances %s in %s', JSON.stringify(i), region);
```
jonasfj: Also not super important, but you might want to know that, both `debug` and `console.log` supports `%j` for json:
```js
debug('killing instances %j in %s', i, region);
```
So no need to stringify :)

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+  var p = Promise.all([
+    that.ec2.describeInstances({
+      Filters: [{
+        Name: 'key-name',
+        Values: [that.keyPrefix + '*']
+      },{
+        Name: 'instance-state-name',
+        Values: ['running', 'pending']
+      }
+    ]}),
```
jonasfj: I hate to point this out, but `describeInstances` may return an `NextToken` property, as well as it make take this property as input. Basically the API requires paging.

The old provisioner doesn't do this, and I suspect that the option is only there because AWS has a really old API.
But the property is there, and EC2 APIs are notoriously slow, so we better implement it... And no you can't really implement this with multi-region-aws-sdk. You need to do per region paging...

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+      reservation.Instances.forEach(function(instance) {
+        var workerType = instance.KeyName.substr(that.keyPrefix.length);
+        x(workerType);
+        rState[workerType][instance.State.Name].push(instance); 
+      });
+    });
+
+    spotReqs[region].SpotInstanceRequests.forEach(function(request) {
+      var workerType = request.LaunchSpecification.KeyName.substr(that.keyPrefix.length);
+    });
```
jonasfj: Aren't you supposed to call `x(workerType)` and some more with `rState`?
This doesn't have any side effects...

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+    }); 
+
+  });
+
+  return state;
+};
+
+
+/**
+ * Get Api State Only
```
jonasfj: Docs about what this returns... depending on what you give it would be nice...

Also shouldn't it be called `getState`...

Note, docs is pretty good in general, I love how it was clear what `_classify` did.

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+  } else if (!region && !type) {
+    return this.__internalState;
+  }
+};
+
+
+/**
+ * List the types known in a given region
+ */
+AwsManager.prototype.typesForRegion = function(region) {
```
jonasfj: rename to `workerTypesForRegion`... If I understand it right... I thought this was instance types.. needless to say I was confused :)

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+
+  return workerTypes;
+};
+
+
+/**
+ * Return a list of all running Instance Ids that are known in this AWS State
+ * These are not categorized by region.  It's one list of strings.
+ */
+AwsManager.prototype.listRunningInstanceIds = function() {
```
jonasfj: Unused member...

jonasfj: Nvm, it's clear what these unused functions do...
And I can see that using `listSpotUnfulfilledRequestIds` to check for old spot requests to cancel because we never expect them to finish is a good idea.

provisioner/aws-manager.js
```diff
@@ -0,0 +1,696 @@
...
+/**
+ * Because the AWS is eventually consistent, it will sometimes take time for
+ * spot requests to show up in the describeSpotInstanceRequests calls for
+ * AWS state.  We will maintain an internal table of these submitted but
+ * not yet visible spot requests so that we can offset the count of a given
+ * instance type for figuring out running capacity.  If the provisioning
+ * process is restarted before the spot request shows up in the api's
+ * state we will lose track of it until it turns into an instance.
+ */
+AwsManager.prototype.trackNewSpotRequest = function(sr) {
```
jonasfj: This is so much more awesome than last time :)

We should doc what an `sr` actually is.. the it what we submitted to EC2, or?

Btw, `reconcileInternalState` is really good name, love it :)

jonasfj: nit, but we should mark this methods with `_`, so it's obvious that they are internal...

provisioner/data.js
```diff
@@ -1,30 +1,615 @@
...
+WorkerType.prototype.determineCapacityChange = function(runningCapacity, pendingCapacity, pending) {
+  assert(typeof runningCapacity === 'number');
+  assert(typeof pendingCapacity === 'number');
+  assert(typeof pending === 'number');
+
+  // We need to know how many total capacity units are extant
+  var totalCapacity = pendingCapacity + runningCapacity;
+
+  // We need to know the current ratio of capacity to pending
+  var percentPending = 1 + pending / totalCapacity;
```
jonasfj: I'm not sure what the `1 + ` is there for... if you're trying to avoid `0/0` you need parentheses...

I suspect we need to do this differently... We need to support `scalingRatio = 0` meaning that when there is more than 0% pending tasks (ie any pending tasks) we should scale up.

provisioner/data.js
```diff
@@ -1,30 +1,615 @@
...
+
+    var regions = Object.keys(that.regions);
+
+    regions.forEach(function(region) {
+      var zones = pricing.__zoneInfo[region];
+      types.forEach(function(type) {
+        zones.forEach(function(zone) {
+          var potentialBid = pricingData[region][type][zone];
+          var potentialPrice = uf[type] * potentialBid;
+          if (!cheapestPrice || potentialPrice < cheapestPrice && potentialPrice < that.maxSpotBid) {
```
jonasfj: Seems to me that you implemented `maxSpotBid` exactly as desired...

provisioner/data.js
```diff
@@ -1,30 +1,615 @@
...
+    } else {
+      throw new Error('Counld not create a bid which satisfies requirements');
+    }
+
+    // This is a sanity check to prevent a screw up where we theoretically
+    // bid $6000 for a spot node.  Code above should make sure that the optimal
+    // bid is selected.  I would argue that if we start bidding on $20/h machines
+    // that we really ought to be very well aware of this, and having to make a
+    // change to the provisioner is a demonstration of our knowledge of that.
+    if (spotBid > 20) {
```
jonasfj: I'm okay lowering this to `10`, but it is really nice to have!! :)

provisioner/data.js
```diff
@@ -1,30 +1,615 @@
...
+    var cheapestPrice;
+    var cheapestRegion;
+    var cheapestZone;
+    var spotBid;
+
+    // Utility Factors, by instance type
+    var uf = {};
+    
+    var types = Object.keys(that.types).map(function(type) {
+      uf[type] = that.utilityOfType(type) || 1;
```
jonasfj: We shouldn't do hardcoded defaults...

provisioner/data.js
jonasfj: I'll push a draft of what I think `scalingRatio` means... Along with an explanation of how it relates to average task runtime.

provisioner/data.js
```diff
@@ -1,30 +1,615 @@
...
+ */
+WorkerType.prototype.determineSpotBids = function(regions, pricing, runningCapacity, pendingCapacity, pending) {
+  assert(regions);
+  assert(pricing);
+  assert(typeof runningCapacity === 'number');
+  assert(typeof pendingCapacity === 'number');
+  assert(typeof pending === 'number');
+  var that = this;
+  
+  var change = this.determineCapacityChange(runningCapacity, pendingCapacity, pending);
```
jonasfj: The change can be negative, in which case we should kill existing spot requests...
It doesn't cost us anything of significance to cancel spot requests. In fact this is all the current provisioner does... It requests/cancels spot requests based on `change` where `change = pendingTasks - pendingNodes`.

jonasfj: See #9

jonasfj:
I addressed a bit of the `scalingRatio` logic issues... If you have a clear coupling from your previous logic to task run time, or expected task wait time, I'm listening...

Either way, if all pending tasks are claimed in one go, we should cancel pending spot requests. If some of them are claimed and no new pending tasks shows up we should cancel pending capacity accordingly (granted we can't cancel capacity that haven't shown up yet.

My original proposal actually contained a scaledown ratio too. Which was supposed to be used to determine when to cancel spot requests, to avoid alternating between request/cancel spot requests... But we can fix that later... The current provisioner sucks more than this, when this is able to cancel excess spot requests.

---
I suggest you go ahead with gradual deployment once this is done. Should hopefully be possible to start that tomorrow, with comments addressed you have an r+ :)
We can take down the old provisioner thursday, when you have migrated all workerTypes away from it. I suggest we do it at the end of our meeting...

bin/provisioner.js
```diff
@@ -1 +1,95 @@
...
+
+  var influx = new base.stats.Influx({
+    connectionString:   cfg.get('influx:connectionString'),
+    maxDelay:           cfg.get('influx:maxDelay'),
+    maxPendingPoints:   cfg.get('influx:maxPendingPoints')
+  });
+
+  var WorkerType = data.WorkerType.setup({
+    table: cfg.get('provisioner:workerTypeTableName'),
+    credentials: cfg.get('azure'),
```
jonasfj: Oh, pretty sure I've said this before... But let's use auth.tc.net for delegating access to the azure account.
It's much easier as you don't have put azure credentials in env vars... just the tc creds, see:
https://github.com/taskcluster/taskcluster-index/blob/master/bin/server.js#L38-L41

In this case:
```js
  var WorkerType = data.WorkerType.setup({
    account: cfg.get('provisioner:azureAccount').
    table: cfg.get('provisioner:workerTypeTableName'),
    credentials: cfg.get('taskcluster'),
    authBaseUrl:      cfg.get('taskcluster:authBaseUrl'),
    context: {
      keyPrefix: keyPrefix,
      provisionerId: provisionerId,
    },
  });
```
`authBaseUrl` is not strictly required, but it's nice... If you do this have you have sufficient scopes you just configure `provisioner:azureAccount` to `taskclusterdev` when testing... and to `taskcluster` in production and you're good to go...

Note, you need to change this in `server.js` too... fix configs and env vars loaded...

routes/v1.js
jhford: I'm well within a 120 limit.  Now that I have eslint running, this would be easy to fix as a follow up sometime.

schemas/get-worker-type-response.yml
jhford: They are subtly different.  The get has a workerType property where the create doesn't have this property.  I guess I could have the property in both, but make an assertion that the `:workerType` param === the one from the body of the document...

I'll keep this in mind as something to fix in a future iteration.

For now, I have rewritten the schema in a much more precise way.

routes/v1.js
jhford: Fixed, simple enough that I just did it.

provisioner/aws-manager.js
jhford: Well, it could be done per-region in multi-region-promised-aws by doing the per-region paging there, but that transforms that library from a dumb wrapper into something with logic and I'm not sure if we want that.

provisioner/aws-manager.js
jhford: Yes!  I found this with the linting stuff as well, but i don't know what happened here...

provisioner/aws-manager.js
jhford: Like this:
```javascript
/**
 * Classify the state received from AWS into something in the shape:
 * {
 *   region: {
 *     workerTypeName: {
 *       running: [<Instance>],
 *       pending: [<Instance>],
 *       spotReq: [<SpotRequest>],
 *     }
 *   }
 * }
 * The Instance and SpotRequest objects are those returned by AWS.
 * We flatten the Reservations because we don't really care about that
 * feature right now.
 */
AwsManager.prototype._classify = function(instanceState, spotReqs) {
```

provisioner/aws-manager.js
jhford: i can see how that's confusing... renamed

provisioner/aws-manager.js
jhford: Done!  Thanks!

provisioner/data.js
jhford: lowered

bin/provisioner.js
jhford: This is something we can do as a follow up pretty easily

schemas/get-worker-type-response.yml
jonasfj: Technically you don't need the `workerType` property in either... But it's nice to have... so I'll leave that decision to you... Making them use the same schema is neat though, it reduces bugs :)

bin/provisioner.js
jonasfj: Yes, but it would be nice if the production azure table credentials only ever exists on auth.taskcluster.net.

We'll probably setup a storage account dedicated to the services that only stores small things.. I have one now called `taskclusterv1`. We could use secondary credentials and then revoke them later, but it just sucks to copy and send credentials around, and having to reset them later... Let's fix initially, it's quicker than emailing them around.

PR #9, opened 2015-03-17
Jhford rewrite jonasfj patches

jonasfj:
A few suggestions... still needs some refactor around `determineSpotBids`; and definitely needs to be able to cancel pending if `change` is negative..

@jhford does this make any sense?

### DIFF
--- a/provisioner/data.js
+++ b/provisioner/data.js
@@ -35,11 +35,12 @@ var WorkerType = base.Entity.configure({
     minCapacity: base.Entity.types.Number,
     /* This is the maximum capacity which we will ever run */
     maxCapacity: base.Entity.types.Number,
-    /* Scaling ratio a ratio of pending jobs to capacity.  A number
-     * which is between 0 and 1 will ensure that there is always 
-     * idle capacity and a number greater than 1 will ensure that
-     * some percentage of tasks will remain pending before we spawn
-     * instances. */
+    /* Scaling ratio a ratio of pending jobs to capacity.
+     * The provisioner will aim to keep `runningCapacity * scalingRatio` number
+     * of pending tasks around. A scaling ratio of 0.2 is equivalent to keeping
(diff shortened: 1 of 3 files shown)


provisioner/data.js
```diff
@@ -529,6 +521,10 @@ WorkerType.prototype.determineSpotBids = function(regions, pricing, runningCapac
   
   var change = this.determineCapacityChange(runningCapacity, pendingCapacity, pending);
 
+  // TODO: If change is negative we should cancel spot requests... This should
+  // go through the aws-manager. IMO this method belongs in the provision.js
+  // file...
```
jonasfj: If we do this... then we're compat with existing provisioner when `scalingRatio = 0` and should roll this out...

I suggest you move `determineSpotBids` into `provsion.js` and add the support for cancelling spot requests.

We may have iterations where we alternate between requesting and canceling spot requests.. Which IMO is fine... We can optimize by prefering nodes that has capacity that isn't above the change we want. Ie. reducing utility by the capacity to desiredChange ratio. But that's not important for now...

Get cancel spot requests done right... this is something we do while running. if `change < 0`.
Then roll this out, it looks really good in general!

jonasfj: The old provisioner does it here:
https://github.com/taskcluster/aws-provisioner/blob/master/provisioner/provision.js#L19-L56

Decide how many to cancel and how many to request... The `chance` returned from `determineCapacityChange` in this PR is already corrected for max/min...

jonasfj:
@jhford, didn't copy in most of these changes a long time ago in:
https://github.com/taskcluster/aws-provisioner/commit/735327f94473781116c53dc48712684cbbd894ae

And replated commits?

If so, let's close this PR...

jhford:
Yes, they are in commit 735327f94473781116c53dc48712684cbbd894ae

Closing this PR

PR #10, opened 2015-03-18
Backport hack for old provisioner

jonasfj:
The old provisioner is really sketchy... so let's avoid testing it who knows what running tests might delete :)

@lightsofapollo, if you think this looks reasonably sane...


lightsofapollo-taskcluster:
@jonasfj lgtm

PR #13, opened 2015-04-28
Jhford rewrite

jhford:
This adds the remaining pulse exchanges with the update messages going to the correct exchange.  This also switches the remainder of the `routes/v1.js` file to using the `async`/`await` syntax.

Not sure how to update the documentation that's posted on docs.tc.net


jonasfj:
Documentation posted to docs.tc.net will be updated when you deploy this to a place that has the environment variable `provisioner_publishMetaData='true'` I configured this for taskcluster-aws-provisioner2 yesterday. And fixed issues it had...

routes/v1.js
```diff
@@ -284,34 +241,29 @@ api.declare({
...
-}, function(req, res) {
+}, async function(req, res) {
   var workerType = req.params.workerType;
 
   if(!req.satisfies({workerType: workerType})) { return undefined; }
 
-  var p = this.WorkerType.load({workerType: workerType});
+  try {
+    var wType = await this.WorkerType.load({workerType: workerType});
+    await wType.remove();
```
jonasfj: This is fine... Performance is hardly important. But it would be faster to use:
https://github.com/taskcluster/taskcluster-base/blob/cd8b589137be779d2c039ede767f9f4a24ef1863/lib/entity.js#L842-L850

jonasfj:
Looks good to me... land it...

PR #15, opened 2015-04-28
Update aws state

jonasfj:
Ugly hack... but at least the UI will be usable...

I'll play around with the idea of refactoring `AwsManager`... might happen, might not, don't block on it!


jhford:
There is something seriously wacky with your branches @jonasfj :)  I'm going to cherry pick the commits over

jonasfj:
oh, it did this again... I'll reset my base :) sorry...

PR #17, opened 2015-04-30
Expire spot requests after 20 minutes

jhford:
EC2 has the facility to automatically close spot requests that exceed a given time range.

Per the docs:

    ValidUntil — (Date)
    
    The end date of the request. If this is a one-time request, the request remains active until 
    all instances launch, the request is canceled, or this date is reached. If the request is 
    persistent, it remains active until it is canceled or this date and time is reached.
    
    Default: The request is effective indefinitely.

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -103,6 +103,8 @@ AwsManager.prototype._filterSpotRequests = function(spotReqs) {
     data.good[region] = [];
     data.stalled[region] = [];
 
+    var now = new Date();
+
     spotReqs[region].SpotInstanceRequests.forEach(function(sr) {
       // These are states which have a state of 'open' but which
       // are likely not to be fulfilled expeditiously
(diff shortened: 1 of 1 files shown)


jhford:
I'm not 100% sure what this option does when the ValidUntil time is hit.  I'm curious if it cancels a pending spot request or if it will kill associated instances.

jhford:
Different approach, @jonasfj , what do you think?

jonasfj:
I seriously doubt that `ValidUntil` will kill the associated instances. We might want to keep our eyes open but that's all..
As docs says,
> "the request remains active until all instances launch .... or this date and time is reached" 
Hence it seems sane to assume, that `ValidUntil` only has the effect of rendering the spot request invalid. Which is also achieved when a node is assigned.

jonasfj:
I think we can land this...

@jhford,
But how about we cancel stalled spot requests after 25 min, and set "ValidUntil: now() + 20min" ?

That way we let AWS deal with most of the expiration.

Note: We should watch after rolling out a potential "ValidUntil" change, as AWS may choose not to provision anything because our ValidUntil is < 1 hour... but don't think it'll be a problem.

jonasfj:
`r+` if any doubt...

jhford:
I merged this!

PR #18, opened 2015-04-30
improvements to cheapest zone/type calculations and randomize regions

jhford:
Improve the bidding logic and log more information about the bidding process.  This prints out a lot more information about the bids that we're creating to get insight into why the provisioner picks a specific bid that it does.


petemoore:
This looks good to me. ++ for the extra logging

PR #19, opened 2015-04-30
Assume that when an instance type is no longer in the workerType definition that it has a capacity of 1 to be safe

jhford:
So the scenario this is guarding is when we update a workerType to remove a workerType.  When we try to look at state to see how much running and pending capacity we have, we need to be able to count those instances which are no longer in the workerType definition.

While assume capacity 1 is not perfect, it errs on the side of safety (over provisioning) and ensures that instances are counted.  We also want instances to kill themselves when the workerType definitions are updated, so overall this feels like a reasonable complexity compromise.

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -387,21 +387,37 @@ AwsManager.prototype.capacityForType = function(workerType, states) {
 
     var wState = rState[wName];
 
+    /* When counting capacity of each type, if we have instances
+       of a type which is no longer in the workerType definition
+       we cannot know what it's capacity is.  We'll try to get its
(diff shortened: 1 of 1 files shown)


petemoore:
I'm going to rubber stamp this due to the urgency, I don't have deep knowledge here.

jonasfj:
As this is merged I took my time to look at it...
Ideally, I would rather ignore instances with illegal instanceType or region, and then kill them...

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1159965
For further discussion... There is some valid point to just letting the configuration changes slowly propagate...

PR #20, opened 2015-05-06
Refactor to lists

jhford:
Didn't actually make a pull request last night... Please don't merge this.

### DIFF
--- a/bin/manage.js
+++ b/bin/manage.js
@@ -82,7 +82,8 @@ function createClient() {
 program
   .version(pkgData.version || 'unknown')
   .description('Perform various management tasks for the Taskcluster AWS Provisioner')
-  .option('-u, --url [url]', 'URL for the API to work against', 'https://taskcluster-aws-provisioner2.herokuapp.com/v1');
+  .option('-u, --url [url]', 'URL for the API to work against', 'http://localhost:5557/v1');
+  //.option('-u, --url [url]', 'URL for the API to work against', 'https://taskcluster-aws-provisioner2.herokuapp.com/v1');
 
 
 program
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -420,6 +420,10 @@ AwsManager.prototype.ensureTags = function() {
         Resources: tags[region][workerType].ids,
       });
 
+      p = p.then(function() {
+        debug('tagged %s/%s: %j', region, workerType, tags[region][workerType].ids);
+      });
+
       // Creating a tag is on best effort basis
       p = p.catch(function(err) {
         debug('Failed to tag %s/%s: %j', region, workerType, tags[region][workerType].ids);
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -5,6 +5,7 @@ var lodash = require('lodash');
 var assert = require('assert');
 var debug = require('debug')('aws-provisioner:aws-manager');
 var objFilter = require('../lib/objFilter');
+var shuffle = require('knuth-shuffle');
 
 /**
  * AWS EC2 state at a specific moment in time
@@ -16,9 +17,12 @@ function AwsManager(ec2, keyPrefix, pubKey) {
   this.ec2 = ec2;
   this.keyPrefix = keyPrefix;
   this.pubKey = pubKey;
-  this.__apiState = {};
-  this.__internalState = {};
   this.__knownKeyPairs = [];
+  this.__apiState = {
+    instances: [],
+    requests: [],
+  };
+  this.__internalState = [];
 }
 
 module.exports = AwsManager;
@@ -152,6 +156,7 @@ AwsManager.prototype.filters = {
     'LaunchTime',
     'Placement:AvailabilityZone',
     'SpotInstanceRequestId',
+    'State:Name',
   ],
 };
 
@@ -172,39 +177,34 @@ AwsManager.prototype.filters = {
  */
 AwsManager.prototype._classify = function(instanceState, spotReqs) {
   var that = this;
-  var state = {};
+
+  /* In the new world, we're just going to store a pair of lists.
+     One list for all intances regardless of region or state and
+     one list for all spot requests regardless of region or state.
+     These lists will be filtered when we need per-region or per-
+     workerType lists */
+  var state = {
+    instances: [],
+    requests: [],
+  };
 
   that.ec2.regions.forEach(function(region) {
-    var rState = state[region] = {};
-
-    function x(type) {
-      if (!rState[type]) {
-        rState[type] = {
-          running: [],
-          pending: [],
-          spotReq: [],
-        };
-      }
-    }
-
     instanceState[region].Reservations.forEach(function(reservation) {
       reservation.Instances.forEach(function(instance) {
         var workerType = instance.KeyName.substr(that.keyPrefix.length);
-        x(workerType);
         var filtered = objFilter(instance, that.filters.instance);
         filtered.Region = region;
         filtered.WorkerType = workerType;
-        rState[workerType][instance.State.Name].push(filtered);
+        state.instances.push(filtered);
       });
     });
 
     spotReqs[region].forEach(function(request) {
       var workerType = request.LaunchSpecification.KeyName.substr(that.keyPrefix.length);
-      x(workerType);
       var filtered = objFilter(request, that.filters.spotReq);
       filtered.Region = region;
       filtered.WorkerType = workerType;
-      rState[workerType].spotReq.push(filtered);
+      state.requests.push(filtered);
     });
 
   });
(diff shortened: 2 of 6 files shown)


bin/manage.js
```diff
@@ -82,7 +82,8 @@ function createClient() {
 program
   .version(pkgData.version || 'unknown')
   .description('Perform various management tasks for the Taskcluster AWS Provisioner')
-  .option('-u, --url [url]', 'URL for the API to work against', 'https://taskcluster-aws-provisioner2.herokuapp.com/v1');
+  .option('-u, --url [url]', 'URL for the API to work against', 'http://localhost:5557/v1');
+  //.option('-u, --url [url]', 'URL for the API to work against', 'https://taskcluster-aws-provisioner2.herokuapp.com/v1');
```
jonasfj: I assume not intentional... How about an option `--localhost` that overwrites `--url`, the port for localhost is hardcoded in the configs anyways...

provisioner/aws-manager.js
```diff
@@ -129,6 +134,32 @@ AwsManager.prototype._filterSpotRequests = function(spotReqs) {
   return data;
 };
 
+
+/**
+ * Instead of defining these lists in lots of places, let's ensure
+ * that we use the same filters in all places!
+ */
+
+AwsManager.prototype.filters = {
```
jonasfj: So as I understand this is a the list of properties on objects representing spot requests and instances, respectively... Is that correct?

One of the things I've been missing was a well-defined representation of these objects. This is definitely a decent solution to that :)
If we go down this road we should probably annotate which of these properties are optional.

I was thinking of doing this with classes, example:
```js
class Instance {
  constructor(resultFromEC2) {
    this.__raw = resultFromEC2
  }
  get instanceId() {
    return this.__raw.InstanceId;
  }
  get availabilityZone() {
    if (!this.__raw.LaunchSpecification ||
        this.__raw.LaunchSpecification.Placement) {
      return undefined;  
    }
    return this.__raw.LaunchSpecification.Placement.AvailabilityZone;
  }
}
```
This way we could flatten the objects and make the properties more accessible. Hide some of the stupid logic that avoids issues when `LaunchSpecification.Placement` is missing. Note this was just a illustrative example, I'm not sure this pattern is better... It certainly adds a lot of unnecessary code.

**Main point** is let's have something that describes what properties the objects have. So we can quickly look it up. We don't have anything like that prior to this PR, and we shouldn't trust AWS docs :)

----
Any reason why these should be accessible outside the class?
Are they not internal.. static? Ie. how about a global constant `const spotReqFilter = [...]`.

provisioner/aws-manager.js
```diff
@@ -15,9 +17,12 @@ function AwsManager(ec2, keyPrefix, pubKey) {
...
   this.keyPrefix = keyPrefix;
   this.pubKey = pubKey;
-  this.__apiState = {};
-  this.__internalState = {};
   this.__knownKeyPairs = [];
+  this.__apiState = {
+    instances: [],
+    requests: [],
+  };
+  this.__internalState = [];
```
jonasfj: I believe I said this in a bug comment too... But let's call it `__submittedSpotReq` or something like that, I'm sure you can give it a better name than me.

My thinking is that it's a spot request that is _submitted_ by not yet visible, so saying it's in the state _submitted_ makes some sense.

provisioner/aws-manager.js
```diff
@@ -180,183 +212,99 @@ AwsManager.prototype._classify = function(instanceState, spotReqs) {
...
-        workerTypes.push(workerType);
-      }
-    });
+/** Return a list of all SpotRequests for a workerType */
+AwsManager.prototype.requestsOfType = function(workerType) {
+  if (typeof workerType === 'string') {
+    workerType = [workerType];
+  }
+  return this.__apiState.requests.filter(function(request) {
+    debugger;
```
jonasfj: probably a left over...

provisioner/aws-manager.js
```diff
@@ -369,84 +317,48 @@ AwsManager.prototype.listSpotRequestIds = function() {
...
-          capacity += 1;
-        }
-      });
+  instances.forEach(function(instance) {
+    if (states.includes(instance.State.Name)) {
+      try {
+        capacity += workerType.capacityOfType(instance.InstanceType);
+      } catch(err) {
+        capacity++;
+      }
```
jonasfj: Another approach using lodash `_.sum` that I just discovered today:
```js
var instanceToCapacity = (instance) => {
  try {
     return workerType.capacityOfType(instance.InstanceType);
  }
  catch(err) {
    return 1;
  }
};

capacity += _.sum(
  instances.filter(instance => {
    return states.includes(instance.State.Name);
  }).map(instanceToCapacity)
);
```

Either way works... But I would consider moving error handling code into `workerType.capacityOfType`.
Or make sure that illegal instanceTypes are ignored and immediately murdered by rougeKiller (seems fair).

provisioner/aws-manager.js
```diff
@@ -494,29 +400,33 @@ AwsManager.prototype._trackNewSpotRequest = function(sr) {
...
+  this.__internalState = this.__internalState.filter(function(request) {
+    // We want to print out some info!
+    if (allKnownSrIds.includes(request.request.SpotInstanceRequestId)) {
+      // Now that it's shown up, we'll remove it from the internal state
+      debug('Spot request %s for %s/%s/%s took %d seconds to show up in API',
+            request.request.SpotInstanceRequestId, request.request.Region,
+            request.request.LaunchSpecification.Placement.AvailabilityZone,
+            request.request.LaunchSpecification.InstanceType,
+            (now - request.submitted) / 1000);
+      return false;
```
jonasfj: `return (now - request.submitted) < 15 * 60 * 1000`

provisioner/aws-manager.js
```diff
@@ -494,29 +400,33 @@ AwsManager.prototype._trackNewSpotRequest = function(sr) {
...
+    } else {
+      debug('Spot request %s for %s/%s/%s still not in api after %d seconds',
+            request.request.SpotInstanceRequestId, request.request.Region,
+            request.request.LaunchSpecification.Placement.AvailabilityZone,
+            request.request.LaunchSpecification.InstanceType,
+            (now - request.submitted) / 1000);
+      // Since we don't have it in the API state yet, we need to keep it in the list
+      return true;
```
jonasfj: return (now - request.submitted) < 15 * 60 * 1000

jonasfj:
My comments here are just nits... I have nothing seriously wrong... This looks pretty neat...

I do wish to encourage flattening the objects representing spot requests and EC2 instances... Even if we just keep them as JSON objects. Following reasons:
 1. No need to check if `Placement` is present before doing `Placement.AvailabilityZone`
 2. Easier to access
 3. Ability rename into more JS-like names (clearly distinguishing raw from internal representation)
 4. Ability to clearly outline which properties we keep in our representation
 5. Easier to filter

Example of utility method I would implement:
```js
AwsManager.prototype.instances = function(predicate) {
  if (!predicate) {
    predicate = () => { return true; }; // Return all if predicate is missing
  } else if (!(predicate instanceof Function)) {
    var conditions = predicate;
    predicate = (instance) => {
      for(var key in conditions) {
        var value = conditions[key];
        if (value instanceof Array) {
           return value.includes(instance[key]);
        }
        return instance[key] === value;
      }
      return true;
    };
  }
  return this.__apiState.instances.filter(predicate);
};

// Usage example:
this.instances({
  workerType:    workerType.workerType,
  stateName:     states
}).forEach(instance => {
   // ....
})

// Instead of:
var instances = this.instancesOfType(workerType.workerType);
instances.forEach(function(instance) {
    if (states.includes(instance.State.Name)) {
      // ....
    }
})
```
You have a lot of places where you do `filter` and `myArray.includes(...)`. By flattening properties this pattern would make a lot of these queries very readable.

bin/manage.js
jhford: It was intentional because I do a lot of testing with delete-all command and I really want to make it extra hard to screw up production.  I'm fine with anything that does not make running on the real server the default :)

provisioner/aws-manager.js
jhford: My line of thinking is that I have __apiState and __internalState.  __submittedSpotReq to me doesn't make it any more clear.  I think adding a comment here is probably the better option.

provisioner/aws-manager.js
jhford: I purposely don't have any sort of abstraction on the api objects as I don't really want to maintain an abstraction on top of the ec2 api.  I'm relying on the API to return properly structured objects, which it does in every case that I've tried.  We actually will get exceptions thrown in the method that reads the Instances and Requests in if they do not contain all of the keys that we want, so I'm not too concerned about fields being missing.

As for storing the filters as a global constant, sure.  I don't see a reason to not expose them.

provisioner/aws-manager.js
jhford: yes! it is

jhford:
I like your idea of the .instances and .requests methods, I'll implement that!

jonasfj:
Note that  `.instances` ad `.requests` would be harder to do... if we don't flatten the objects...
As you'll have to make the `predicate` function a bit more complicated to deal with nested values.
I'm not exactly sure how to do that...

Maybe:
```js
this.instances({
   'State.Name':   states
}).forEach(....
```

Would work... Note that `'State.Name'` works with `_.get` and `_.property` two neat lodash utilities..

provisioner/aws-manager.js
jhford: Killing them immediately doesn't feel right to me.  If we have 1000 r3.xlarge jobs which are underway but we want to switch to r3.2xlarge *only* because we're not getting xlarge ones anymore, why would we want to kill those 1000 jobs?  That puts us 1000 capacity behind, wastes the nodes we have and feels wrong.  I think that in this case, I'd rather that the worker receive the workerType updated message as we discussed and handle killing itself.

I'd like to minimize the number of moving pieces in this code for this refactor as I'd like to make sure that it gets deployed before my development head gets too far from production.  We can definately look at this later!

provisioner/aws-manager.js
jhford: yep!  good call!

jhford:
Opening a new pull request so to avoid confusion.

jhford:
nvm, since it's not merged that actually makes it even more confusing!

jhford:
@jonasfj ok, i'm feeling pretty happy about this branch.  I'd like to merge it to avoid drifting too far from production.  I tested the old-state-format-emulation with a localhost copy of taskcluster-tools and it all works!

provisioner/aws-manager.js
jonasfj: > Killing them immediately doesn't feel right to me. If we have 1000 r3.xlarge jobs which are underway but we want to switch to r3.2xlarge only because we're not getting xlarge ones anymore, why would we want to kill those 1000 jobs? 

In this scenario we shouldn't remove `r3.xlarge` from the list of allowed instance types. If we remove `r3.xlarge` it **must** be because we've deemed the instanceType incapable of reliably running the tasks we want to run and the speed we want to run them. Or something like that, ie. the instanceType is no longer allowed.

Well, at least that is one view of it :)

---
On topic, I'm okay with us solving this in [bug 1159965](https://bugzilla.mozilla.org/show_bug.cgi?id=1159965).

For the purpose of this refactor it certainly doesn't block.

jonasfj:
`r+` this is generally ready to land...

Some of the comments should be trivial to address. At least the review comments on todays reviews, they are mostly ignore or copy in the code/docs-comment I wrote.

You landed a lot of awesome features very quickly here, I completely support landing it :)

jhford:
This was merged days ago, but I forgot to close this PR, so doing now!

PR #22, opened 2015-05-11
bug 1159965 - AwsManager.prototype.rougeKiller needs to kill illegal instanceTypes

jhford:
Assume capacity === 1 when we can't find it.  This is as elsewhere in the code.

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -825,7 +825,11 @@ AwsManager.prototype.killCapacityOfWorkerType = function(workerType, count, stat
 
   // Build the mapping of capacity to instance type string
   workerType.instanceTypes.forEach(function(t) {
-    caps[t.instanceType] = workerType.capacityOfType(t.instanceType);
+    try {
+      caps[t.instanceType] = workerType.capacityOfType(t.instanceType);
+    } catch (err) {
+      caps[t.instanceType] = 1;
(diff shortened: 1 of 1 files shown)


jonasfj:
I don't see this fixing bug 1159965, but I now agree with you that we should fix that bug.
And instead address the issue of updates through bug [1145688](https://bugzilla.mozilla.org/show_bug.cgi?id=1145688).

Do merge this... :)

provisioner/aws-manager.js
```diff
@@ -825,7 +825,11 @@ AwsManager.prototype.killCapacityOfWorkerType = function(workerType, count, stat
 
   // Build the mapping of capacity to instance type string
   workerType.instanceTypes.forEach(function(t) {
-    caps[t.instanceType] = workerType.capacityOfType(t.instanceType);
+    try {
+      caps[t.instanceType] = workerType.capacityOfType(t.instanceType);
+    } catch (err) {
+      caps[t.instanceType] = 1;
+    }
```
jonasfj: I see this try, catch around `capacityOfType` alot... any chance we should just add it inside the `capacityOfType` method?

jhford: I want to make sure that the caller understands that they are doing something with undefined behaviour instead of silently returning 1 or undefined.

PR #23, opened 2015-05-13
Stats

jhford:
Aside from 518896a, these are the first set of stats things.  For the stats, I'm thinking that I should put all the series into a single module instead of defining them in the constructor of the aws manager...

I'm not done adding stats, these are just the first sets.  Please let me know how they look to you.

One question that I do have is how I should query influx later on.... Is it easier to add to base.stats the ability to do queries or should I use something else for that?

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var Promise = require('promise');
-var lodash = require('lodash');
 var assert = require('assert');
 var debug = require('debug')('aws-provisioner:aws-manager');
 var objFilter = require('../lib/objFilter');
@@ -12,7 +11,7 @@ var base = require('taskcluster-base');
 /**
  * AWS EC2 state at a specific moment in time
  */
-function AwsManager(ec2, provisionerId, keyPrefix, pubKey, maxInstanceLife, influx) {
+function AwsManager (ec2, provisionerId, keyPrefix, pubKey, maxInstanceLife, influx) {
   assert(ec2);
   assert(provisionerId);
   assert(keyPrefix);
@@ -79,7 +78,7 @@ function AwsManager(ec2, provisionerId, keyPrefix, pubKey, maxInstanceLife, infl
       az: base.stats.types.String,
       instanceType: base.stats.types.String,
       workerType: base.stats.types.String,
-    }
+    },
   });
 }
 
@@ -89,7 +88,7 @@ module.exports = AwsManager;
  * Update the state from the AWS API and return a promise
  * with no resolution value when completed.
  */
-AwsManager.prototype.update = function() {
+AwsManager.prototype.update = function () {
   var that = this;
 
   var p = Promise.all([
@@ -118,13 +117,13 @@ AwsManager.prototype.update = function() {
     }),
   ]);
 
-  p = p.then(function(res) {
+  p = p.then(function (res) {
     var filteredSpotRequests = that._filterSpotRequests(res[1]);
     that.__apiState = that._classify(res[0], filteredSpotRequests.good);
     return that.handleStalledRequests(filteredSpotRequests.stalled);
   });
 
-  p = p.then(function() {
+  p = p.then(function () {
     // We want to make sure that our internal state is always up to date when
     // we fetch the updated state
     that._reconcileInternalState();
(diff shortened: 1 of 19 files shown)


jhford:
Oh, please don't merge this yet...

Regarding the linter stuff:  I think it's 100% critical to use a linter and strict mode personally.  I value consistency over things looking 100% pretty, which is why I have such a strict eslintrc file.  I think the value of linting outweighs the ugliness it adds.

Regarding the unit tests, we talked about the issues with the capacity change method, so until we have a fix, i'll leave those uncommented.  Is there a way in mocha to mark a test as expected fail but still run it -- without inverting the assertions?

provisioner/aws-manager.js
```diff
@@ -1,33 +1,85 @@
...
+      id: base.stats.types.String,
+      // other columns should be obvious.
+      // This column is 0 for it showed up somehow, somewhere
+      // and 1 for being dropped on the floor
+      didShow: base.stats.types.Number,
+      // How many seconds to show up in API.  This is a maximum
+      // bound since we only check the API once every iteration
+      lag: base.stats.types.Number,
+    },
+  });
```
jonasfj: My recommendation is to use `base.stats.Series` as a type constructor. Ie. do this in the global scope, as follows:
```js
// Global scope
var Ec2ApiLagSeries = new base.stats.Series({
  name: 'Ec2ApiLag', // Perhaps prefix it like: "Provisioner.Ec2ApiLag" or something
  columns: {....}
});

function AwsManager(...., influx) {
   ....
   this.reportEc2ApiLag = Ec2ApiLagSeries.reporter(influx);
};

AwsManager.prototype._reconcileInternalState = function () {
  ....
  that.reportEc2ApiLag({ // You can't missspell series, and if a property is missing it will complain 
    region: request.request.Region,
    az: request.request.LaunchSpecification.Placement.AvailabilityZone,
    ...
  });
};
```

Note, since all of these are provisioner specific measures.. Maybe we should prefix them `AwsProvisioner.` or something...
Also I would add `provisionerId` as a column. In influxdb 0.9 string columns will be tags, or at least we'll convert some of them to tags. So they'll be cheap (which they aren't now).

provisioner/data.js
```diff
@@ -625,7 +646,8 @@ WorkerType.prototype.determineSpotBids = function(managedRegions, pricing, chang
     if (cheapestBid && cheapestPrice <= that.maxPrice) {
       change -= that.capacityOfType(cheapestType);
       spotBids.push({
-        price: cheapestBid,
+        price: cheapestBid, // Ugh, awful naming!
+        truePrice: cheapestPrice, // for history reasons
```
jonasfj: he he, refactor later I assume... :)

test/test_workertype.js
```diff
@@ -0,0 +1,519 @@
...
+  filename: 'taskcluster-aws-provisioner',
+});
+
+var keyPrefix = cfg.get('provisioner:awsKeyPrefix');
+//var pubKey = cfg.get('provisioner:awsInstancePubkey');
+var provisionerId = cfg.get('provisioner:id');
+//var maxInstanceLife = cfg.get('provisioner:maxInstanceLife');
+
+// We don't use influx yet, but may as well include it
+/* var influx = new base.stats.Influx({
```
jonasfj: For testing you can also use the `base.stats.NullDrain`. it has a method `pendingPoints` so you can get a count of how many pending properties there is...

jonasfj:
I skipped most of the lint changes... but the other commits seems solid.

jonasfj:
> Regarding the linter stuff: I think it's 100% critical to use a linter and strict mode personally. I value consistency over things looking 100% pretty, which is why I have such a strict eslintrc file. I think the value of linting outweighs the ugliness it adds.

I think it's a good idea to play with this... I'm open to investigating this some more, and possibly rolling it out on all of TC components if we can come up with a good set rules :)

The big sell to me would be http://eslint.org/docs/rules/key-spacing
```js
// "key-spacing": [2, { "align": "value" }]
// beforeColon and afterColon default to false and true, respectively
var obj = {
    a:    value,
    bcde: 42,

    fg: foo(),
    h:  function() {
        return this.a;
    },
    ijkl: 'Non-consecutive lines form a new group'
};
```
If I could be allowed to make properties aligned when it's reasonably possible I would be totally into using a linter :)
@jhford, I support whatever the provision does, but let's open a discussion about linter somewhere else... Feel free to file a bug...

But strict mode us unimportant when we preprocess with babeljs (it will force strict mode when es6/7 features are used).
IMO we should also preprocess with `flowtype`, well maybe someday :)

> Regarding the unit tests, we talked about the issues with the capacity change method, so until we have a fix, i'll leave those uncommented. Is there a way in mocha to mark a test as expected fail but still run it -- without inverting the assertions?

Sounds good to me.

provisioner/aws-manager.js
jhford: cool.  i've done this.  I was trying to figure out how the reporter mechanism worked, and I thought I did what you listed above but it just didn't work.  I now have the reporter stuff working and I've verified that!

provisioner/data.js
jhford: yes!  ugh, naming is annoying.

jhford:
Aligned properties is something that I personally strongly dislike, but I'd prefer to have lint rules than none, so i guess we can figure something out.

I've fixed the issues here, so I'm going to look into pushing this branch live!

jonasfj:
@jhford,
Sorry, to clarify: You do not have to align properties :)
As long as the linter doesn't complain when I do it... That's all I care about. I don't want to force my bad habits on everybody!

I agree that there are many cases where it's not a good idea (it certainly messes up diffs).
But there is also places where it truly improves readability, at least for me. I often use JSON objects as sort tables.

PR #24, opened 2015-06-02
Stats for provisioner

jhford:
Submit relelvant stats regarding provisioning to influx.

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -11,6 +11,16 @@ var series = require('./influx-series');
 
 const MAX_ITERATIONS_FOR_STATE_RESOLUTION = 20;
 
+function dateForInflux(thingy) {
+  if (typeof thingy === 'object' && thingy.getTime) {
+    // assume this is a date object
+    return thingy.getTime();
+  } else if (typeof thingy === 'string')  {
+    return new Date(thingy).getTime();
+  }
+  throw new Error('dont know how to thing this thingy');
+}
+
 /**
  * AWS EC2 state at a specific moment in time
  */
@@ -130,13 +140,15 @@ AwsManager.prototype.update = function () {
     // TODO: Probably better to do a single set of describe calls and bucket
     //       them in this section... hmm, no strong feelings until data
 
-    // Because we don't filter the spot requests, we'll need to make sure
-    // that we remove the SpotInstanceRequests per-region response so that
+    // Because we don't filter the spot requests, we'll need to make sure that
+    // we remove the SpotInstanceRequests per-region response so that
     // that._classify() knows how to interpret the data
     Object.keys(res[3]).forEach(function(region) {
       deadSpotRequests[region] = res[3][region].SpotInstanceRequests;
     });
 
+    // We don't currently handle this case because it seems to be deprecated in
+    // the docs.  We should be alerted if it begins to be used again
     if (livingInstances.NextToken && livingInstances.NextToken !== '') {
       console.log('[alert-operator] WARNING!  We have a ' +
           'DescribeInstances NextToken and arent using it!!!');
@@ -148,15 +160,17 @@ AwsManager.prototype.update = function () {
     that.__apiState = that._classify(livingInstances, filteredSpotRequests.good);
     that.__deadState = that._classify(deadInstances, deadSpotRequests);
 
-    var stateDifferences = that._compareStates(that.__apiState, that.__previousApiState);
-
-    console.log('Differences:');
-    console.log(JSON.stringify(stateDifferences, null, 2));
-
-    /*console.log('Dead state:');
-    console.log(JSON.stringify(that.__deadState, null, 2));*/
-
-    that._reconcileStateDifferences(stateDifferences, that.__deadState, that.__apiState);
+    // remove this try/catch once this has been live for a while
+    try {
+      var stateDifferences = that._compareStates(that.__apiState, that.__previousApiState, that.__deadState);
+      that._reconcileStateDifferences(stateDifferences, that.__deadState, that.__apiState);
+    } catch(err) {
+      console.log('[alert-operator] failure in computing differences');
+      console.log(err);
+      if (err.stack) {
+        console.log(err.stack);
+      }
+    }
 
     return that.handleStalledRequests(filteredSpotRequests.stalled);
   });
@@ -237,7 +251,7 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
           region: region,
           az: sr.LaunchSpecification.Placement.AvailabilityZone,
           instanceType: sr.LaunchSpecification.InstanceType,
-          time: new Date().toISOString(),
+          time: dateForInflux(new Date()),
           price: parseFloat(sr.SpotPrice, 10),
           reason: 'spot-request-price-too-low',
         });
@@ -266,41 +280,52 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
  * state.  You'll need to have another data source to find the resolution of
  * the now missing resources
  */
-AwsManager.prototype._compareStates = function (newState, previousState) {
-  var nowMissing = {
+AwsManager.prototype._compareStates = function (newState, previousState, deadState) {
+  assert(newState);
+  assert(previousState);
+  assert(deadState);
+
+  var missingIds = {
     instances: [],
     requests: [],
   };
 
+  // to make comparison of states easier, we create a list of all the ids of
+  // both spot requests and instances in each of the two compared states
+  var allInstancesInPreviousState = previousState.instances.map(function(instance) {
+    return instance.InstanceId;
+  });
+  var allRequestsInPreviousState = previousState.requests.map(function(request) {
+    return request.SpotInstanceRequestId;
+  });
   var allInstancesInNewState = newState.instances.map(function(instance) {
     return instance.InstanceId;
   });
-
   var allRequestsInNewState = newState.requests.map(function(request) {
     return request.SpotInstanceRequestId;
   });
 
-  previousState.instances.forEach(function (instance) {
-    var id = instance.InstanceId;
-    if (allInstancesInNewState.includes(id)) {
-      debug('instance %s has not terminated yet', id);
-    } else {
-      debug('instance %s has terminated', id);
-      nowMissing.instances.push(instance);
-    }
+  // find all the instances and request ids which were in the previous state
+  // but not in the new state
+  missingIds.instances = allInstancesInPreviousState.filter(function(id) {
+    return !allInstancesInNewState.includes(id);
+  });
+  missingIds.requests = allRequestsInPreviousState.filter(function(id) {
+    return !allRequestsInNewState.includes(id);
   });
 
-  previousState.requests.forEach(function (request) {
-    var id = request.SpotInstanceRequestId;
-    if (allRequestsInNewState.includes(id)) {
-      debug('spot request %s is still unfulfilled', id);
-    } else {
-      debug('spot request %s is no longer open', id);
-      nowMissing.requests.push(request);
-    }
-  });
-
-  return nowMissing;
+  // Now let's grab those instances and requests which are absent, but instead
+  // let's use their new state object instead of the old one.  This is to avoid
+  // the problem of getting the stale state info in the later methods which
+  // need information about why the state change occured
+  return {
+    instances: deadState.instances.filter(function(instance) {
+      return missingIds.instances.includes(instance.InstanceId);
+    }),
+    requests: deadState.requests.filter(function(request) {
+      return missingIds.requests.includes(request.SpotInstanceRequestId);
+    }),
+  };
 };
 
 
@@ -324,12 +349,12 @@ AwsManager.prototype._reconcileStateDifferences = function(differences, deadStat
     that.reportSpotRequestsFulfilled({
       provisionerId: that.provisionerId,
       region: request.Region,
-      az: request.Placement.AvailabilityZone,
-      instanceType: request.InstanceType,
+      az: request.LaunchSpecification.Placement.AvailabilityZone,
+      instanceType: request.LaunchSpecification.InstanceType,
       workerType: request.WorkerType,
       id: request.SpotInstanceRequestId,
       instanceId: request.InstanceId,
-      time: request.Status.UpdateTime,
+      time: dateForInflux(request.Status.UpdateTime),
     });
     debug('spot request %j fulfilled!', request);
   }
@@ -347,7 +372,7 @@ AwsManager.prototype._reconcileStateDifferences = function(differences, deadStat
       // do not have an api state which reflects that
       that.__awaitingSpotFulfilmentStatus.push({
         id: request.SpotInstanceRequestId,
-        time: new Date().toISOString(),
+        time: new Date(),
         iterationCount: 0,
       });
     } else {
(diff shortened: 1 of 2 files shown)


jhford:
a couple things:

1. I'm pretty sure that we don't need to handle the case where an instance/request does not yet have a state change reason.... what do you think?
1. all of the codepaths are tested for me except the case where we have an instance get spot-price killed submitting to the spot price floor table.  I have the whole stats stuff wrapped in a try-catch and will remove the try-catch when I've seen this code working reliably.

provisioner/influx-series.js
```diff
@@ -39,6 +39,78 @@ module.exports.spotRequestsSubmitted = new base.stats.Series({
   },
 });
 
+// Store the times when a spot request is fulfilled
+module.exports.spotRequestsFulfilled = new base.stats.Series({
```
jonasfj: How about just doing: `exports.spotRequestsFulfilled` that goes to the entire file. Or use es6 style export like `export var spotRequestsSubmitted = ...` or even
```js
var spotRequestsFulfilled = ...

export {spotRequestsSubmitted, spotRequestsFulfilled}
```
Also `'use strict'` is not needed with babeljs.

Just two random things.. `module.exports` is a bit disturbing here.

provisioner/influx-series.js
```diff
@@ -39,6 +39,78 @@ module.exports.spotRequestsSubmitted = new base.stats.Series({
...
+  name: 'AwsProvisioner.SpotRequestsFulfilled',
+  columns: {
+    provisionerId: base.stats.types.String,
+    region: base.stats.types.String,
+    az: base.stats.types.String,
+    instanceType: base.stats.types.String,
+    workerType: base.stats.types.String,
+    id: base.stats.types.String,
+    instanceId: base.stats.types.String,
```
jonasfj: Do we really want `id` and `instanceId` ?
These won't translate to tags in 0.9 because they have a much higher cardinality.

That said I still think we can have them... They are just slow to query by, takes up a lot of space and is hard to use. But if we have use-cases it probably not a problem.

provisioner/influx-series.js
```diff
@@ -39,6 +39,78 @@ module.exports.spotRequestsSubmitted = new base.stats.Series({
...
+    instanceType: base.stats.types.String,
+    workerType: base.stats.types.String,
+    id: base.stats.types.String,
+    // This is the true time per the ec2 api
+    time: base.stats.types.Number,
+    bid: base.stats.types.Number,
+    state: base.stats.types.String,
+    // These are the status fields
+    statusCode: base.stats.types.String,
+    statusMsg: base.stats.types.String,
```
jonasfj: Same argument with `id` as above... I think `statusMsg` might be something we want to store. Or maybe we would just rather print it to the log... I'm we'll never really aggregate it...

Anyways, same argument as with `id`, do we need it and do we have use cases for this.

Note, as we probably don't have so many `spotRequestsDied` entries whatever might be fine :)

provisioner/influx-series.js
jonasfj: This discussion is relevant to everything... but assuming we don't make 1k spot reqs a minutes, the load impossed by this data is probably not important.. So leaving the strings is probably a sound approach to this :)

provisioner/aws-manager.js
```diff
@@ -3,11 +3,23 @@
...
-var objFilter = require('../lib/objFilter');
+// XXX var objFilter = require('../lib/objFilter');
 var shuffle = require('knuth-shuffle');
 var taskcluster = require('taskcluster-client');
 var series = require('./influx-series');
 
+const MAX_ITERATIONS_FOR_STATE_RESOLUTION = 20;
+
+function dateForInflux (thingy) {
+  if (typeof thingy === 'object' && thingy.getTime) {
```
jonasfj: `instanceof Date` is a good way to check :)
(Not that it matters)

provisioner/aws-manager.js
```diff
@@ -3,11 +3,23 @@
...
+    // assume this is a date object
+    return thingy.getTime();
+  } else if (typeof thingy === 'string') {
+    return new Date(thingy).getTime();
```
jonasfj: Oh, if you want to be clear safe, you can check the parsing with `isFinite(new Date(thingy))` if not finite the parse failed and the date object is `NaN`. Probably overkill.

provisioner/aws-manager.js
```diff
@@ -70,15 +102,75 @@ AwsManager.prototype.update = function () {
...
+    that.ec2.describeSpotInstanceRequests({
+      Filters: [
+        {
+          Name: 'launch.key-name',
+          Values: [that.keyPrefix + '*'],
+        }, {
+          Name: 'state',
+          Values: ['cancelled', 'failed', 'closed', 'active'],
+        },
+      ],
```
jonasfj: How about making two requests? and then filter them afterwards... That is likely much more efficient.

-----
**Unrelated detail**, do consider using `await/async`, you can do things like `var [instances, spotRequests] = await Promise.all([...` (destructive array assignments rock with `Promise.all`)
You'll be so happy, just write `AwsManager.prototype.update = async function` and it works.

provisioner/aws-manager.js
```diff
@@ -507,12 +865,11 @@ AwsManager.prototype._trackNewSpotRequest = function (sr) {
...
   assert(sr);
 
-  var that = this;
-
   var allKnownSrIds = this.knownSpotInstanceRequestIds();
 
   if (!allKnownSrIds.includes(sr.request.SpotInstanceRequestId)) {
-    var filtered = objFilter(sr.request, that.filters.spotReq);
+    // XXX var filtered = objFilter(sr.request, that.filters.spotReq);
+    var filtered = sr.request;
```
jonasfj: I'm curious, why drop the filtering? The filtering is about removing properties right?

provisioner/aws-manager.js
jonasfj: I see you have todo longer down... :)

That seems fine for now...

provisioner/aws-manager.js
```diff
@@ -70,15 +102,75 @@ AwsManager.prototype.update = function () {
...
+
+    // TODO: Probably better to do a single set of describe calls and bucket
+    //       them in this section... hmm, no strong feelings until data
+
+    // Because we don't filter the spot requests, we'll need to make sure that
+    // we remove the SpotInstanceRequests per-region response so that
+    // that._classify() knows how to interpret the data
+    Object.keys(res[3]).forEach(function (region) {
+      deadSpotRequests[region] = res[3][region].SpotInstanceRequests;
+    });
```
jonasfj: Curious, why not filter them? If we need additional properties how about we just extend the filter to keep those...
I think it's preferable to have objects for spot requests that are similar form (has same set of properties).

provisioner/aws-manager.js
```diff
@@ -70,15 +102,75 @@ AwsManager.prototype.update = function () {
...
+    } catch(err) {
+      console.log('[alert-operator] failure in computing differences');
+      console.log(err);
+      if (err.stack) {
+        console.log(err.stack);
+      }
+    }
+
     return that.handleStalledRequests(filteredSpotRequests.stalled);
   });
```
jonasfj: Don't know exactly where to put this comment. But we should use information about dead spot requests in `_reconcileInternalState`, so that it's even less likely that we have to fallback and rely on the spot request timeout thing...

provisioner/aws-manager.js
```diff
@@ -70,15 +102,75 @@ AwsManager.prototype.update = function () {
...
-    };
-    var filteredSpotRequests = that._filterSpotRequests(res[1]);
-    that.__apiState = that._classify(res[0], filteredSpotRequests.good);
+    }
+    var filteredSpotRequests = that._filterSpotRequests(livingSpotRequests);
+
+    // We store the last iteration's state
+    that.__previousApiState = that.__apiState;
+    that.__apiState = that._classify(livingInstances, filteredSpotRequests.good);
+    that.__deadState = that._classify(deadInstances, deadSpotRequests);
```
jonasfj: I'm wondering, why have two sets of state? ie. four lists in total, isn't it tempting to just filter by state every time you query... If adding the `.filter` method I suggested a while ago...

provisioner/aws-manager.js
```diff
@@ -166,6 +271,254 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
...
+    that instance has an associated state transition reason.  For when there is a
+    spot price termination, we should also look for a spot request in the good and
+    dead ones to see what we bid on the thing.  We should also track what the time
+    was when we found the difference and we should only try for up to, say, 5 times
+    before we give up on trying to figure out why the instance terminated.
+
+    http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_StateReason.html
+  */
+
+  function plotInstanceDeath (instance, time) {
```
jonasfj: Please start using ES6 arrow functions, there is no reason not to...

**example**
```js
let plotInstanceDeath = (instance, time) => {
  this.reportInstanceTerminated(....
};
```

Notice that you don't have to use the `that` thing, when you use arrow functions...

provisioner/aws-manager.js
```diff
@@ -166,6 +271,254 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
...
+        });
+      } else {
+        debug('Could not find a price for a spot-price killed instance');
+      }
+    }
+  }
+
+  // Let's handle instance which already have a state reason, or
+  // save them for a future iteration if not
+  differences.instances.forEach(function (instance) {
```
jonasfj: again, another great place to start using arrow functions... So you don't have to write `that`.
**example**
```js
differences.instances.forEach(instance => {
  ...
  this...
});
```
Notice that because there is only a single parameter you can write it as `param => {/* body */}`.

provisioner/aws-manager.js
```diff
@@ -166,6 +271,254 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
...
+        iterationCount: 0,
+      });
+    }
+  });
+
+  // Now, let's try to account for those instance which are awaiting a state reason
+  this.__awaitingStateReason = this.__awaitingStateReason.filter(function (instanceAwaiting) {
+    var keepItInTheList = true;
+    deadState.instances.forEach(function (instanceMightHave) {
+      if (instanceMightHave.InstanceId === instanceAwaiting.id) {
```
jonasfj: Fyi, conditions like this can also be done as:
```js
deadState.instances.filter(i => i.InstanceId === instanceAwaiting.id)
                   .forEach(instanceMightHave => {
   ...
})
```
Anyways, it's a matter of taste... Just wanted to give an idea...

provisioner/aws-manager.js
```diff
@@ -166,6 +271,254 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
...
+      debug('spot request %j did something else', request);
+    }
+  });
+
+  this.__awaitingSpotFulfilmentStatus = this.__awaitingSpotFulfilmentStatus.filter(function (requestAwaiting) {
+    var keepInTheList = true;
+
+    deadState.requests.forEach(function (requestMightHave) {
+      if (requestMightHave.SpotInstanceRequestId === requestAwaiting) {
+        if (requestMightHave.State === 'active' && requestMightHave.Status.Code === 'filfilled') {
```
jonasfj: Is `filfilled` correct?

provisioner/aws-manager.js
```diff
@@ -166,6 +271,254 @@ AwsManager.prototype._filterSpotRequests = function (spotReqs) {
...
+      });
+      debug('spot request %j did something else', request);
+    }
+  });
+
+  this.__awaitingSpotFulfilmentStatus = this.__awaitingSpotFulfilmentStatus.filter(function (requestAwaiting) {
+    var keepInTheList = true;
+
+    deadState.requests.forEach(function (requestMightHave) {
+      if (requestMightHave.SpotInstanceRequestId === requestAwaiting) {
```
jonasfj: Hmm, have you considered using `_.find(deadState.requests, r => r.SpotInstanceRequestId === requestAwaiting);`

If there is non. `_.find` will return `null`. You can dramatically reduce number of indentation.. and 
When you reach `keepInTheList = false` you can just `return false` immediately instead, example:
```js
this.__awaitingSpotFulfilmentStatus = this.__awaitingSpotFulfilmentStatus.filter(requestAwaiting => {
  // Look for spot request, and report it if furfilled
  var spotReq = _.find(deadState.requests, r => r.SpotInstanceRequestId === requestAwaiting);
  if (spotReq && spotReq.State === 'active' && spotReq.Status.Code === 'furfilled') {
    plotSpotFulfilment(spotReq);
    return false; // Remove from list, when reported.
  }

  return requestAwaiting.iterationCount++ <= MAX_ITERATIONS_FOR_STATE_RESOLUTION;
});
```

jonasfj:
So `provisioner/aws-manager.js` is pretty complex... it's just getting even harder to follow with double the amount of lists. I can't really see if it works, but I don't see anything critically wrong with it either...

You have `r+` to land this, as I think it'll work, if you search the code for the place that says`filfilled`, you probably want it to say `furfilled` (just a guess). But I recommend that you refactor some of the code to be nicer and apply babeljs features (arrow function, destructive assignment, await/async, etc.).

------
Note, I would strongly advice that you look at ES6 + await/async, and start rewriting some of this... And keep it cleaner... You can choose which of my comments to ignore or apply, some of them are a matter of taste other dramatically improves code quality.

provisioner/aws-manager.js
jhford: I only need to work on the dead state in one place, but the live state everywhere.  joining the lists has no memory usage improvement and ensures that in a bunch of places we'll need to either add a second filter, or complicate existing filters.  Maybe if we do another major internal structure refactoring, we could deal with it then, but for now I just don't see that as a high priority

provisioner/aws-manager.js
jhford: Good idea.  I don't want to block on it, but I do like the idea a lot... https://bugzilla.mozilla.org/show_bug.cgi?id=1171003

provisioner/aws-manager.js
jhford: nope, good catch

provisioner/aws-manager.js
jhford: Yes.  I needed a couple new properties, and those are properites that might not be present in yet-unfulfilled spot requests and living instances, which objFilter doesn't currently handle cleanly.  I'll be turning this back on if memory usage increases dramatically.

provisioner/influx-series.js
jhford: module.exports === exports, i thought.  I've always done module.exports...

use strict is just habbit and doesn't hurt i guess.

provisioner/influx-series.js
jhford: I'd rather have important data there with slow provisioner related queries than not to have the data when we're trying to figure out why something is going wrong.  As for the Ids and instance Ids, we need *something* to relate a log/treeherder page to a datapoint in influx in my opinion, instance id and spot request id can do that.

jhford:
I do want to do the await/async stuff and arrow functions, I'm just not in the mindset of doing them by default.  It's muscle memory winning out :/  I do agree that I need to do a pass and convert this file over to async/await.

I'm going to move forward with deployment!

provisioner/influx-series.js
jonasfj: `module.exports === exports` true... and it is a matter of taste... `exports` is just easier to read...

Not important...

PR #25, opened 2015-06-03
add and test an api end point to track when a worker comes online

jhford:
Quick endpoint with a test for checking in a workerType.  I don't know that we'll ever use it in production, but having it available is part of my Q2 goals and I'd like to have it here until the secrets stuff is a lot further along

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -9,9 +9,10 @@ var AwsManager = require('../provisioner/aws-manager');
 var v1 = require('../routes/v1');
 var Aws = require('multi-region-promised-aws');
 var _ = require('lodash');
+var series = require('../provisioner/influx-series');
 
 /** Launch server */
-var launch = function (profile) {
+var launch = function (profile, useNullInfluxDrain) {
   // Load configuration
   var cfg = base.config({
     defaults: require('../config/defaults'),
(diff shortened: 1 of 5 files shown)


bin/server.js
```diff
@@ -41,11 +42,23 @@ var launch = function (profile) {
...
-    maxPendingPoints: cfg.get('influx:maxPendingPoints'),
-  });
+  var influx;
+  if (useNullInfluxDrain) {
+    var influx = new base.stats.NullDrain();
+
+    // Ugly*1000 but otherwise, it'll be really hard to get a reference to this
+    // drain aiui, which we need to get the number of pending points to make
+    // sure that the stat was submitted for real
+    module.exports.influx = influx;
```
jonasfj: Don't use globals unless strictly necessary, you can set it on the `server` object created by `app.createServer();`.

bin/server.js
```diff
@@ -9,9 +9,10 @@ var AwsManager = require('../provisioner/aws-manager');
 var v1 = require('../routes/v1');
 var Aws = require('multi-region-promised-aws');
 var _ = require('lodash');
+var series = require('../provisioner/influx-series');
 
 /** Launch server */
-var launch = function (profile) {
+var launch = function (profile, useNullInfluxDrain) {
```
jonasfj: Why not use `config.get('influx') === null` instead, and then disable influxdb in `config/test.js`.

test/helper.js
```diff
@@ -53,7 +53,9 @@ mocha.before(async () => {
     clients: defaultClients,
   });
 
-  webServer = await bin.server('test');
+  webServer = await bin.server('test', true);
+  // Ugh, again pretty ugly
+  exports.influx = bin.server.influx;
```
jonasfj: Why not take it from `webServer` which you can add the property to in `server.js`.

jonasfj:
Yeah, I see no reason to adopt this before we have secrets, which will probably use a different callback...

It looks okay, though I would definitely find an alternative way to expose `useNullInfluxDrain`.
 1. Probably set `influx = null` in `config/test.js` and use that to indicate that a `NullDrain` should be used
 2. Use `server` object from `app.createServer();` instead of using a global, it's very ugly.

jonasfj:
So, you can land this if you do (1) and (2), but I don't see any reason why we would update workers to do a best-effort callback if we're going to add secrets later anyways.

Note, secrets should be quick to add... Mostly it's just about issues temporary creds with some predetermined scopes and somes scopes from the workerType definition. The rest is static hardcoded secrets..

PR #26, opened 2015-06-09
Secrets

jhford:
still need to fix the data/extra stuff we talked about but here's a PR to review.


provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    minPrice: base.Entity.types.Number,
+    maxPrice: base.Entity.types.Number,
+    canUseOndemand: base.Entity.types.JSON, // delete this
+    canUseSpot: base.Entity.types.JSON, // delete this
+    instanceTypes: base.Entity.types.JSON,
+    regions: base.Entity.types.JSON,
+    // Store the date of last modification for this entity
+    lastModified: base.Entity.types.Date,
+    // Global base UserData object for overwriting
+    userData: base.Entity.types.JSON,
```
jonasfj: yay no more AWS JSON, with capital U in `UserData`.

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    // Global base LaunchSpecification object for overwriting
+    launchSpec: base.Entity.types.JSON,
+    // Global base secrets object for overwriting
+    secrets: base.Entity.types.JSON,
+    // Global base scope list for appending, these are the scopes that
+    // temporary TC credentials will be issued against
+    scopes: base.Entity.types.JSON,
+  },
+  migrate: function (item) {
```
jonasfj: Why not do this ES6 style..
```js
  },
  migrate(item) {
    ...
  }
```

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    console.log('Upgrading a workerType to version 2');
+    console.log(JSON.stringify(item, null, 2));
+    var oldUserData;
+    if (item.launchSpecification.UserData) {
+      oldUserData = JSON.parse(new Buffer(item.launchSpecification.UserData, 'base64').toString());
+    } else {
+      oldUserData = {};
+    }
```
jonasfj: Drop the `else` clause:
```js
let oldUserData = {};
if (item.launchSpecification.UserData) {
  oldUserData = JSON.parse(new Buffer(item.launchSpecification.UserData, 'base64').toString());
}
```

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    var oldLaunchSpec = item.launchSpecification;
+    delete item.launchSpecification;
```
jonasfj: I haven't tested modifying the existing object... It should just be a flat object, and I think it does work.
But it's not part of any of the examples or tests from `taskcluster-base`, [see docs](https://github.com/taskcluster/taskcluster-base/blob/26cd5e3b824f52016440512ccea5f954b774f26c/lib/entity.js#L211-L216).

If you want to use this pattern, please add a test doing it to `taskcluster-base`, or just make a fresh object, I think it's safer...

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    delete oldLaunchSpec.UserData;
+    item.launchSpec = lodash.clone(oldLaunchSpec);
+    item.secrets = {};
+    item.scopes = [];
+    item.userData = lodash.clone(oldUserData);
```
jonasfj: You should probably do a `_.cloneDeep` in both places...

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+
+    // Now, strip out the UserData and LaunchSpec stuff from existing entries
+    // and put them in the right structure then create empty scopes and secrets
+    item.regions.forEach(r => {
```
jonasfj: I feel like modifying existing object is sketchy... or at least hard to read...
How about `item.regions = item.region.map(r => {....`
Then return a new `{userData:, scopes:, secrets: ,....}` object...

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+      // Let's get the old launch spec and user data
+      var oldUserData;
+      if (r.overwrites && r.overwrites.UserData) {
+        oldUserData = JSON.parse(new Buffer(r.overwrites.UserData, 'base64').toString());
+        ['capacity', 'provisionerId', 'workerType'].forEach(key => {
+          if (oldUserData[key]) {
+            delete oldUserData[key];
+          }
+        });
+      } else {
```
jonasfj: Again drop `else` set `oldUserData = {}` initially.

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+    item.userData = lodash.clone(oldUserData);
+
+    // Now, strip out the UserData and LaunchSpec stuff from existing entries
+    // and put them in the right structure then create empty scopes and secrets
+    item.regions.forEach(r => {
+      // Let's get the old launch spec and user data
+      var oldUserData;
+      if (r.overwrites && r.overwrites.UserData) {
+        oldUserData = JSON.parse(new Buffer(r.overwrites.UserData, 'base64').toString());
+        ['capacity', 'provisionerId', 'workerType'].forEach(key => {
```
jonasfj: `['capacity', 'provisionerId', 'workerType'].forEach(k => delete oldUserData[k]);`
As far as I can see no need to check if it's truthy...

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+
+      // Now set them up
+      r.userData = oldUserData;
+      r.launchSpec = oldOverwrites;
+      r.secrets = {};
+      r.scopes = [];
+
+    });
+
+    item.instanceTypes.forEach(t => {
```
jonasfj: Same transformation as above... let's deduplicate the code..
Option A:
do `_.union(item.regions, item.instanceTypes).forEach(e => { /* transform e */ })`
Option B:
`item.regions = item.regions.map(transformEntry)` and them do the same for `instanceTypes`...

I guess transforming the existing object isn't bad if you do it with option A.

provisioner/data.js
jonasfj: nvm... looking at the call site for `migrate` it's clear that you can modify the existing object and return that.
In fact there is absolutely no reason not to do this... Other than readability, but that is a matter of taste.

You still welcome to ensure the pattern is covered by a test though, I always welcome an extra test :)

provisioner/data.js
```diff
@@ -86,6 +87,107 @@ var WorkerType = base.Entity.configure({
...
+      // Now set them up
+      t.userData = oldUserData;
+      t.launchSpec = oldOverwrites;
+      t.secrets = {};
+      t.scopes = [];
+    });
+
+    item.lastModified = new Date();
+    console.log('===>');
+    console.log(JSON.stringify(item, null, 2));
```
jonasfj: So I recommend that maybe you use `debug` as follows:
```js
var debugMigrate = require('debug')('provisioner:data:migrate');
....
  migrate(item) {
    ....
    debugMigrate("Migrated v1 => v2: %j => %j", oldItem, newItem);
  }
```

This way you can control when you want to see these heavy debug messages. Probably bad in production, but might be nice when debugging locally (sometimes). Oh, and `debug` doesn't do the string interpolation if `provisioner:data:migrate` isn't enabled by the `DEBUG` environment variable, so the cost of this is zero when disabled (`debugMigrate` is just an empty function, when disabled).

provisioner/data.js
jonasfj: I realize that if you wanted to... one could probably use `_.cloneDeep` to make two objects... but yeah, I don't see the need if your transformation is on the existing object, which is fine considering we change few properties.

provisioner/data.js
```diff
@@ -224,32 +327,51 @@ WorkerType.prototype.testLaunchSpecs = function () {
...
-      hasType = true;
+  assert(foundRegion, 'Region for workertype not found');
+
+  // Find the instanceType overwrites object, assert if type is not found
+  var instanceTypeOverwriteObjects = {};
+  var foundInstanceType = false;
+  worker.instanceTypes.forEach((t) => {
+    if (t.instanceType === instanceType) {
+      if (foundInstanceType) {
+        throw new Error('InstanceType must be unique');
```
jonasfj: So many lines and so much indentation for something very simple.
```js
let instanceTypeInfo = _.find(worker.instanceTypes, {instanceType});
assert(instanceTypeInfo, 'InstanceType for workertype not found');
instanceTypeOverwriteObjects = _.defaults({}, instanceTypeInfo, {
  launchSpec: {},
  userData: {},
  secrets: {},
  scopes: []
});
```
Note, I'm not sure if `instanceTypeInfo`  has additional properties, but if it does you can use `_.pick` to project it down. Or you can do lines like `instanceTypeOverwriteObjects.launchSpec = instanceTypeInfo.launchSpec || {}`, that's fine too.. But please use `_.find` rather than iterating the list yourself (it returns null if not found).

Note, the argument `{instanceType}` given to find is ES6 syntax that expands to `{instanceType: instanceType}`. It's just a nice short hand, in the hands of `_.find` it will return the first object where the `instanceType` property equals the value given.

I don't think this is the right place to enforce the invariant `'InstanceType must be unique'`. Here it's better to assume that it is unique. I might consider doing an `[alert-operator]` message if instanceType is missing however.

jonasfj: This goes for `regionOverwriteObjects` too..,

jonasfj: oh, I missed that this was for testing... then you might want to check that you don't have duplicates in `workerTypes`.
For this you can use `_.filter` instead of `_.find` which also takes an object as filter condition...
So you get:
```js
let instanceTypeInfos = _.filter(worker.instanceTypes, {instanceType});
assert(instanceTypeInfos.length === 1, 'Expected exactly one instanceType definition');
instanceTypeOverwriteObjects = _.defaults({}, instanceTypeInfos[0], {
....
```

provisioner/data.js
```diff
@@ -265,83 +387,51 @@ WorkerType.createLaunchSpec = function (region, instanceType, worker, keyPrefix,
...
-
   // Check for type specific keys in the general keys and region keys
   typeSpecificKeys.forEach(function (key) {
-    if (worker.launchSpecification[key]) {
+    if (worker.launchSpec[key]) {
       throw new Error(key + ' is type specific, not general');
     }
-    if (regionOverwrites[key]) {
+    if (regionOverwriteObjects.launchSpec[key]) {
       throw new Error(key + ' is type specific, not type specific');
```
jonasfj: Error message is `region` not `type` twice..

provisioner/data.js
```diff
@@ -265,83 +387,51 @@ WorkerType.createLaunchSpec = function (region, instanceType, worker, keyPrefix,
   var regionSpecificKeys = [
```
jonasfj: Should be made `const` I see no reason not to make it a global `const REGION_SPECIFIC_KEYS = [...];` list.

provisioner/data.js
```diff
@@ -265,83 +387,51 @@ WorkerType.createLaunchSpec = function (region, instanceType, worker, keyPrefix,
...
-    } else {
-      allInstanceTypes.push(i);
-    }
+  var config = {};
+
+  // Do the cascading overwrites of the object things
+  ['launchSpec', 'userData', 'secrets'].forEach(x => {
+    config[x] = lodash.cloneDeep(worker[x] || {});
+    lodash.assign(config[x], regionOverwriteObjects[x]);
+    lodash.assign(config[x], instanceTypeOverwriteObjects[x]);
```
jonasfj: Neat, note that you can give more than one to `_.assign`, hence:
```js
config[x] = _.cloneDeep(_.assign({}, worker[x], regionOverwriteObjects[x], instanceTypeOverwriteObjects[x]));
```
Note, cloneDeep is only need if you actually modify the properties inside `config[x]` later...

provisioner/secret.js
```diff
@@ -0,0 +1,49 @@
...
+var assert = require('assert');
+var debug = require('debug')('aws-provisioner:Secret')
+var taskcluster = require('taskcluster-client');
+
+var Secret = base.Entity.configure({
+  version: 1,
+
+  // TODO: verify these are valid azure keys - heeeelp please Jonas
+  partitionKey: base.Entity.keys.StringKey('token'),
+  rowKey: base.Entity.keys.StringKey('provisionerId'),
```
jonasfj: I would recommend:
```js
partitionKey: base.Entity.keys.StringKey('token'),
rowKey: base.Entity.ConstantKey('secret')
```

Note, I don't see any reason for `provisionerId` different provisioners should use different azure tables, I think that's safest :)

provisioner/secret.js
```diff
@@ -0,0 +1,49 @@
...
+  var properties = {
+    token: token,
+    created: new Date(),
+    provisionerId: provisionerId,
+    workerType: body.workerType,
+    secrets: body.secrets,
+  };
+  debug('Created Secret entity %s: %j', token, body.secrets);
+  return base.Entity.create.call(this, properties);
+};
```
jonasfj: Why change the original API, it's pretty decent already...

provisioner/secret.js
```diff
@@ -0,0 +1,49 @@
...
+  partitionKey: base.Entity.keys.StringKey('token'),
+  rowKey: base.Entity.keys.StringKey('provisionerId'),
+
+  properties: {
+    token: base.Entity.types.String,
+    created: base.Entity.types.Date,
+    provisionerId: base.Entity.types.String,
+    workerType: base.Entity.types.String,
+    secrets: base.Entity.types.JSON,
+  },
```
jonasfj: More than `created` I would like one called `expiry`... Because it is at creation we've decided how long time the spot request is valid for... And this must expire `1 hour` after the spot request has expired...

I don't recall if we do this already, but I'm pretty sure we planned to put an expiration in our spot requests...

provisioner/secret.js
```diff
@@ -0,0 +1,49 @@
...
+  debug('Created Secret entity %s: %j', token, body.secrets);
+  return base.Entity.create.call(this, properties);
+};
+
+Secret.prototype.modify = function() {
+  throw new Error('No modifications to secrets are allowed');
+};
+
+Secret.prototype.listAll = function() {
+  throw new Error();
```
jonasfj: This doesn't even exist...

routes/v1.js
```diff
@@ -335,6 +354,132 @@ api.declare({
...
+    token: token,
+    provisionerId: this.provisionerId,
+  });
+
+  p = p.then(function(secret) {
+    return secret.remove();
+  });
+
+  p = p.then(function() {
+    res.reply({outcome: 'success'});
```
jonasfj: No need for this... Don't return anything but a `204`. So probably `res.status(204).end()` or something like that...

routes/v1.js
```diff
@@ -335,6 +354,132 @@ api.declare({
...
+    'Remove a secret.  It is very important that the consumer of a ',
+    'secret delete the secret from storage before handing over control',
+    'to another process or else it could read the HTTP UserData endpoint',
+    'and use the getSecrete() api here to get the secrets',
+  ].join('\n'),
+}, async function (req, res) {
+  var input = req.body;
+  var token = req.params.token;
+
+  var p = this.Secret.load({
```
jonasfj: No need to load..

Just do `await this.Secret.remove({token, provisionerId}, true)`... This will remove the entity and ignore errors if the entity have been deleted already (perhaps due to retries). See that with `async + await` there is no `p` :)

See docs: https://github.com/taskcluster/taskcluster-base/blob/26cd5e3b824f52016440512ccea5f954b774f26c/lib/entity.js#L747-L755

(if you remove provisioner from rowKey as I suggested you can remove it here too!).

routes/v1.js
```diff
@@ -335,6 +354,132 @@ api.declare({
...
+    return secret.remove();
+  });
+
+  p = p.then(function() {
+    res.reply({outcome: 'success'});
+  });
+
+  p = p.catch(function(err) {
+    errorHandler(err, res, token);
+    return err;
```
jonasfj: We don't want `ResourceNotFound` to be a 404. The operation should be idempotent.

provisioner/secret.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
```
jonasfj: You need a `Secrets.expire` method...
something like this:
```js
Secret.expire = async function(now) {
  assert(now instanceof Date, "now must be given as option");
  var count = 0;
  await base.Entity.scan.call(this, {
    expires: base.Entity.op.lessThan(now)
  }, {
    limit: 250, // max number of concurrent delete operations
    handler: (item) => { count++; return item.remove(true); }
  });
  return count;
};
```

Note, `count` is nice to have for testing... Then you just create a secret and run this with `now = now() + 5 days` and count that at least something was deleted.

jonasfj:
I don't see any major bugs... Some code that could be more elegant... The migrate thing seems to work.. and all you need is to add `expires` and expire old secrets...
How often I don't know... maybe every 15min... as part of the provisioning process, we won't have many secrets alive, so scanning the table is likely quick... (But if you add some stats for it that might be cool).

routes/v1.js
```diff
@@ -335,6 +354,132 @@ api.declare({
...
+  var that = this;
+
+  var p = this.Secret.load({
+    token: token,
+    provisionerId: this.provisionerId,
+  });
+
+  p = p.then(function(secret) {
+    that.reportSecurityTokenRetreived({securityToken: token});
+    return res.reply(secret.secrets);
```
jonasfj: Here we should reply:
```js
{
  data: secret.secrets,
  credentials: taskcluster.createTemporaryCredentials({
    start, expiry,
    scopes: secret.scopes,
    credentials: this.credentials // from config
  })
}
```

provisioner/secret.js
jonasfj: Just realized we need to add a property `scopes` too.. as in the scopes that should be used for the temporary credentials that will be issued when you call to get these secrets.

provisioner/provision.js
```diff
@@ -198,6 +202,37 @@ Provisioner.prototype.runAllProvisionersOnce = function () {
...
+  var launchInfo = workerType.createLaunchSpec(bid.region, bid.type);
+
+  // This should probably move from here to createLaunchSpec but that
+  // can happen later
+  launchInfo.launchSpec.Placement = {
+    AvailabilityZone: bid.zone,
+  };
+
+  // TODO: Generate temporary credentials and stick 'em in around about
+  //       here
```
jonasfj: No, do this when people get the secrets... No need to store the credentials in azure table storage when we can settle with just storing the scopes...

provisioner/data.js
jhford: Because I'm not familar yet with ES6 class notation.  I do intend to switch to it, but I'd like to get this working before I fix style issues :)

provisioner/data.js
jhford: duh, good idea, fixed

provisioner/data.js
jhford: Oh right, a new object is probably the more sane option!!!  Yes, I'll switch to using cloneDeep

jhford: fwiw, all of the migration tests that I could find in taskcluster-base mutate the passed-in object and return it...  If the passed in object is just the entity's __properties object, then modifying/recreating doesn't make much difference I guess

jhford: Oh, remember the issue that I was having were I saw things being migrated after having done a modify?  i just realised that I was modifying the item passed in but not actually returning it...  That could've been the root of my problem!

provisioner/data.js
jhford: done

provisioner/data.js
jhford: if I were deleting r.overwrites, yes, but r.overwrites.UserData.keydoesntexist would throw a TypeError if UserData doesn't exist

provisioner/data.js
jhford: done

provisioner/data.js
jhford: Actually, since i'm now creating a new worker and using map, i can no longer handle instance types and regions the same as they have a slightly different set of properties.

provisioner/data.js
jhford: yep! the intention was to delete this before going into production as one of the cleanups for the final review.

provisioner/provision.js
jhford: interesting thought, I'll do that.

provisioner/secret.js
jhford: yep, originally you had suggested that we use one table with a provisioner id field, but this makes much more sense to me :)

provisioner/secret.js
jhford: I'm pretty sure that we expire unfulfilled spot requests after a certain amount of time, which is not configurable.

provisioner/secret.js
jhford: I was cribbing off the WorkerType entity, you're right in this case, I'll drop this method

jhford:
@jonasfj Hey, so I've fixed up the stuff and implemented the data keys for secrets and userdata.  I'm going to fix the linter stuff this afternoon and if I don't hear anything to the contrary i'll be doing a deploy tomorrow

provisioner/data.js
```diff
@@ -86,6 +88,110 @@ var WorkerType = base.Entity.configure({
...
+        instanceType: t.instanceType,
+        capacity: t.capacity,
+        utility: t.utility,
+        secrets: {},
+        scopes: [],
+      };
+
+      it.userData = fixUserData(t.overwrites.UserData);
+      it.launchSpec = lodash.cloneDeep(t.overwrite);
+      delete it.launchSpec.UserData;
```
jonasfj: Why aren't these part of the construct above?

You don't even need to create `it`, you can just do:
```js
return {
  instanceType: t.instanceType,
  capacity: t.capacity,
  utility: t.utility,
  secrets: {},
  scopes: [],
  userData: fixUserData(t.overwrites.UserData),
  launchSpec: lodash.omit(lodash.cloneDeep(t.overwrite), 'UserData')
};
```

Exactly the same goes to the regions transformations above too..

provisioner/data.js
```diff
@@ -86,6 +88,110 @@ var WorkerType = base.Entity.configure({
...
+      newWorker.userData = fixUserData(item.launchSpecification.UserData);
+    }
+
+    // We want to make sure that the global user data is correct
+    newWorker.launchSpec = lodash.cloneDeep(item.launchSpecification);
+    delete newWorker.launchSpec.UserData;
+
+    // Now we set up defaults for the secrets and scopes
+    newWorker.secrets = {};
+    newWorker.scopes = [];
```
jonasfj: Why not set that in `var newWorker = {` where you quote `set up the static/easy data` :)

Just a nit...

provisioner/data.js
```diff
@@ -86,6 +88,110 @@ var WorkerType = base.Entity.configure({
...
+    }
+
+    // Global User Data
+    if (item.launchSpecification.UserData) {
+      newWorker.userData = fixUserData(item.launchSpecification.UserData);
+    }
+
+    // We want to make sure that the global user data is correct
+    newWorker.launchSpec = lodash.cloneDeep(item.launchSpecification);
+    delete newWorker.launchSpec.UserData;
```
jonasfj: It's a matter of preference, but `lodash.omit` is nice too... And then you can set it as part of the object creation above...

provisioner/data.js
```diff
@@ -86,6 +88,110 @@ var WorkerType = base.Entity.configure({
...
+      // so we don't really need them to ever be in UserData
+      ['provisionerId', 'workerType', 'capacity'].forEach(y => {
+        delete ud[y];
+      });
+      return ud;
+    }
+
+    // Global User Data
+    if (item.launchSpecification.UserData) {
+      newWorker.userData = fixUserData(item.launchSpecification.UserData);
```
jonasfj: So if there is no user data, we just omit the property? shouldn't it be `{}` if there is none... I'm not sure?

If so, it seems that it might be better to make `fixUserData` robust enough to return `{}` when given something that isn't a string.

provisioner/influx-series.js
```diff
@@ -36,6 +36,15 @@ module.exports.spotRequestsSubmitted = new base.stats.Series({
...
     price: base.stats.types.Number,
+    securityToken: base.stats.types.String,
+  },
+});
+
+// Store when a security token was retreived
+module.exports.securityTokenRetreived = new base.stats.Series({
+  name: 'AwsProvisioner.SecurityTokenRetreived',
+  columns: {
+    securityToken: base.stats.types.String,
```
jonasfj: Why?
I don't like this... There is no reason to leak security tokens into influxdb... which has horrible security...
Both here and in `spotRequestsSubmitted`.

---

If you want to record time from spot request submitted to secret retrieved, that's great but we can do that in other ways, but looking at the `created` timestamp on `secret` entities. I realize they don't have one, but adding one shouldn't take many seconds.

provisioner/secret.js
```diff
@@ -0,0 +1,23 @@
...
+  version: 1,
+
+  partitionKey: base.Entity.keys.StringKey('token'),
+  rowKey: base.Entity.keys.ConstantKey('secret'),
+
+  properties: {
+    token: base.Entity.types.String,
+    expiration: base.Entity.types.Date,
+    workerType: base.Entity.types.String,
+    secrets: base.Entity.types.JSON,
```
jonasfj: nit, but considering the name of the property in the response maybe this should be called `data` instead...

Note, I think this is fine.. In some ways it might be interesting to add `created`, `instanceType`, `region` and `availabilityZone` here... If nothing else then just for recording statistics...

routes/v1.js
```diff
@@ -85,8 +86,8 @@ api.declare({
   name: 'createWorkerType',
   deferAuth: true,
   scopes: ['aws-provisioner:manage-worker-type:<workerType>'],
-  input: SCHEMA_PREFIX_CONST + 'create-worker-type-request.json#',
-  output: SCHEMA_PREFIX_CONST + 'get-worker-type-response.json#',
+  input: SCHEMA_PREFIX_CONST + 'create-worker-type2-request.json#',
+  output: SCHEMA_PREFIX_CONST + 'get-worker-type2-response.json#',
```
jonasfj: Do we need the old one still ?

routes/v1.js
```diff
@@ -335,6 +358,141 @@ api.declare({
...
+}, async function (req, res) {
+  var token = req.params.token;
+  var that = this;
+
+  var p = this.Secret.load({
+    token: token,
+  });
+
+  p = p.then(function (secret) {
+    that.reportSecurityTokenRetreived({securityToken: token});
```
jonasfj: bad... leaking secret bits... if the worker doesn't manage to delete secret before it crashes for unknown reason... OR someone races with worker to get secret and delete it...

routes/v1.js
```diff
@@ -335,6 +358,141 @@ api.declare({
...
+      token: token,
+      provisionerId: this.provisionerId,
+    });
+
+    var match = [
+      'workerType',
+      'secrets',
+      'token',
+      'scopes',
+      //'expiration' weird stuff is happening here.  going to assume that
```
jonasfj: `secret.expiration` is a Date object, the other one is a string.

routes/v1.js
```diff
@@ -335,6 +358,141 @@ api.declare({
...
+      // we should probably do some sort of Date.toISOString() comparison or something
+    ].every((key) => {
+      return _.isEqual(secret[key], input[key]);
+    });
```
jonasfj: You only do this because the end-point is only for testing...

jonasfj: nvm... I get this is for idempotency... Awesome :)

jonasfj:
I only see nits.. Apart from leaking secrets into stats.. please don't do that :)

provisioner/influx-series.js
jhford: but we need something to correlate specific instances.  I could alternatively create a different slugid that's used just for this purpose, but we need some sort of unique identifier to be able to link a spot request submitted (which we don't know the ID until after it's submitted) to a worker coming online.

I'll implement a second slugid for this purpose.

jhford:
Ok, i am tracking the spot request id when i submit the spot request.  I guess we could make it mandatory to include a spot request id with the security token, but I almost think that if the endpoint does two different things with two different ids, we should just have two endpoints.  In support of that, I think that the time that the docker worker takes to claim the secrets and set it up isn't really fair to include in the AWS overhead.  The hit to check in should happen as the first or second thing in the worker.

provisioner/influx-series.js
jonasfj: 1) We can't effectively make queries over this that tries to measure time between random ids in two different series... We don't even want to write an influxdb query for this. I assume you want to correlation for stats to measure time between submission and running (call back via secret) right?

2) Note, you create the `secret` entity, about the same time as you submit the spot request, so why not simply avoid correlating the instances, and instead add a `created` timestamp to the `secret` entity... Won't that do?

jonasfj:
> Ok, i am tracking the spot request id when i submit the spot request. I guess we could make it mandatory to include a spot request id with the security token, but I almost think that if the endpoint does two different things with two different ids, we should just have two endpoints.

I agree, if the end-point does two things it should be two different end-points.

> In support of that, I think that the time that the docker worker takes to claim the secrets and set it up isn't really fair to include in the AWS overhead. The hit to check in should happen as the first or second thing in the worker.

Well, before docker-worker starts it does things like format local disks and launch docker daemon.
I'm sure we don't want to build a special service (on the AMI) just to callback for reporting time...

---

Anyways, I think the measures we want here is, and correct me I'm wrong:

A) Time `spot-request-submitted` to `instance-starts-booting`
B) Time from `instance-starts-booting` to `secrets-callback`

(A) is the time it takes to get an instance, and (B) is the time it takes from we got an instance till it's operational, I know secrets callback isn't the same as first task claim, but it's pretty close.
Note, we have to design around (A) because it's a property of AWS; but (B) is something we can improve.

I fear that we can only get the timestamp for `instance-starts-booting` from AWS cli... Using the `SpotRequestId` property on instances it should be possible to find (A).

I'm not sure if we can do (B), or if it's better for docker-worker to report this as part of it's own statistics. It could just use system `uptime` to determine time from boot to launch. Granted this probably doesn't count the time incurred waiting for the AMI to loaded into the VM.

Maybe, (C): Time `spot-request-submitted` to `secrets-callback`
Is a practically measurable duration that approximates well...

If not I suggest a second callback with instanceId.

-----------------------

** Actually** we could correlate these on a best effort basis. But inserting `spotRequestId` in the `secret` entity. We can't have it there initially, because we **must** create the `secret` entity before we create the spot request, otherwise an instance could try to get a secret that isn't there and we loose money (I would rather loose statistics).

Anyways, we just do  a `secret.modify`... I'll add inline comment to demonstrate.

provisioner/provision.js
```diff
@@ -198,6 +203,37 @@ Provisioner.prototype.runAllProvisionersOnce = function () {
...
+    secrets: launchInfo.secrets,
+    scopes: launchInfo.scopes,
+    expiration: taskcluster.fromNow('40 mins'),
+  });
+
+  p = p.then(()=> {
+    this.awsManager.requestSpotInstance(launchInfo, bid);
+  });
+
+  return p;
```
jonasfj: How about:
```js
let secret = await this.Secret.create({
  token: launchInfo.securityToken,
  workerType: workerType.workerType,
  secrets: launchInfo.secrets,
  scopes: launchInfo.scopes,
  expiration: taskcluster.fromNow('40 mins'),
  spotRequestId: '' // Empty string to indicate we don't have any
});

var spotReqId = await this.awsManager.requestSpotInstance(launchInfo, bid);

await secret.modify(secret => {
  secret.spotRequestId = spotReqId;
});
```

Now when we get a callback for the secret, if `secret.spotRequestId !== ''` we look for the spotRequest and submit duration stats... If `secret.spotRequestId == ''` then we don't try to correlate... This will happen very very rarely.

jonasfj: Note, you would need to modify `AwsManager#requestSpotInstance` to return spot request id.

provisioner/influx-series.js
jonasfj: By:
> We don't even want to write an influxdb query for this.

I mean I can't imagine how we would compute durations by comparing ids from different influxdb queries... Isn't that going to cause a lot of subqueries (like one per row)?

PR #27, opened 2015-06-10
awaitified bin/server.js

jonasfj:
I haven't tested this, because I wasn't sure how to run tests (and what I would have to cleanup on ec2)... please have a look this just a simple refactor to async/await.


jhford:
This looks good to me, i will test it out then merge

PR #28, opened 2015-06-23
Changed json entry from type: date to type: string / format: date-time

petemoore:
@jhford 
This would make the date declarations consistent with other dates we have in our schemas. Also this would fix the parsing in taskcluster-client-go which unfortunately doesn't work with a date type. I think this is consistent with the allowed values in http://json-schema.org/latest/json-schema-core.html#anchor2 (see http://tools.ietf.org/html/rfc4627#section-2.1)

### DIFF
--- a/schemas/get-worker-type2-response.yml
+++ b/schemas/get-worker-type2-response.yml
@@ -76,7 +76,8 @@ properties:
       True if this worker type is allowed spot instances.  Currently ignored
       as all instances are Spot
   lastModified:
-    type: date
+    type: string
+    format: date-time
     description: |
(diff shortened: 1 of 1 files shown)


petemoore:
I think the date format we are using elsewhere is e.g. [2016-06-22T18:50:44.564036Z](https://queue.taskcluster.net/v1/task/lzRphoWHQVCUHlne9aQ7sg)

petemoore:
Closing, as this got fixed in fa7d90a629157fdbe250c6a82f10a120d3ff76f8. Thanks John!

PR #30, opened 2015-07-02
improvements to the iteration logic

jhford:
fixes issues with the provisioner getting stuck and not killing itself.

### DIFF
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -13,7 +13,7 @@ module.exports = {
     publishMetaData:      'false',
     statsComponent:       undefined,
     awsKeyPrefix:         undefined,
-    iterationInterval:    1000 * 1,
+    iterationInterval:    1000 * 10,
     awsInstancePubkey:    undefined,
     allowedRegions:       undefined, // comma seperated list
     maxInstanceLife:      '- 96 hours',
--- a/config/defaults.js
+++ b/config/defaults.js
@@ -13,7 +13,7 @@ module.exports = {
     publishMetaData:      'false',
     statsComponent:       undefined,
     awsKeyPrefix:         undefined,
-    iterationInterval:    1000 * 10,
+    iterationInterval:    1000 * 1,
     awsInstancePubkey:    undefined,
     allowedRegions:       undefined, // comma seperated list
     maxInstanceLife:      '- 96 hours',
--- a/provisioner/provision.js
+++ b/provisioner/provision.js
@@ -142,23 +142,24 @@ Provisioner.prototype.run = function () {
     if (this.__keepRunning && !process.env.PROVISION_ONCE) {
       debug('scheduling another iteration in %d seconds', 
         Math.round(this.provisionIterationInterval / 1000));
-      setTimeout(() => {
-        provisionIteration().catch(() => {
-          process.exit(1);
-        });
-      }, this.provisionIterationInterval);
+      sched(this.provisionIterationInterval);
     } else {
       debug('not scheduling further iterations');
     }
   }
 
+  function sched (t) {
+    setTimeout(() => {
+      provisionIteration().catch(err => {
+        debug('[alert-operator] failure! %s %s', err, err.stack); 
+        process.exit(1);
+      });
+    }, t);
+  }
+
   // To ensure that the first iteration is not called differently that
   // subsequent ones, we'll call it with a zero second timeout
-  setTimeout(() => {
-    provisionIteration().catch(() => {
-      process.exit(1);
-    });
-  }, 0);
+  sched(0);
 
 };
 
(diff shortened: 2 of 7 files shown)


provisioner/provision.js
```diff
@@ -79,61 +80,90 @@ var stats = {
...
-    that.__watchDog.touch();
+    this.__watchDog.touch();
 
     // We should make sure that we're not just permanently failing
     // We also don't want to
-    if (stats.iterations > 20 && stats.failures > stats.success) {
-      debug('[alert-operator] killing provisioner because it has run ' +
-            'for a while but has failed lots of iterations');
+    if (stats.failure > MAX_FAILURES) {
+      debug('[alert-operator] dieing after %d failures', MAX_FAILURES);
```
selenamarie: 'dying'

provisioner/provision.js
```diff
@@ -79,61 +80,90 @@ var stats = {
...
 Provisioner.prototype.run = function () {
-  var that = this;
-
   this.__keepRunning = true;
+
   this.__watchDog.on('expired', function () {
-    debug('[alert-operator] Provisioner is stuck, killing');
+    debug('[alert-operator] provisioning iteration exceeded max time');
+    exitTimer(MAX_KILL_TIME);
     throw new Error('WatchDog expired');
```
jonasfj: So probably bubbling up to `setTimeout` didn't do the trick before?

provisioner/provision.js
```diff
@@ -79,61 +80,90 @@ var stats = {
   failure: 0,
 };
```
jonasfj: A global stats object? isn't that a little "interesting"?

provisioner/provision.js
```diff
@@ -79,61 +80,90 @@ var stats = {
...
-    debug('[alert-operator] Provisioner is stuck, killing');
+    debug('[alert-operator] provisioning iteration exceeded max time');
+    exitTimer(MAX_KILL_TIME);
     throw new Error('WatchDog expired');
   });
+
   this.__watchDog.start();
 
-  function provisionIteration () {
+  var provisionIteration = async () => {
```
jonasfj: Okay, I love that you're starting to use `async` one of the really cool things is that you no longer need to make an infinite loop, your code boils down to this:
```
do {
  debug('starting iteration %d, successes %d, failures %d',
          this.iterations,
          this.success,
          thís.failures);
  this.__watchDog.touch();

  // We should make sure that we're not just permanently failing
  // We also don't want to
  if (this.failures > MAX_FAILURES) {
    debug('[alert-operator] dieing after %d failures', MAX_FAILURES);
    exitTimer(MAX_KILL_TIME);
    throw new Error('provisioner is failing a lot');
  }

  this.iterations++;
  try {
    await this.runAllProvisionersOnce();
    stats.success++;
  } catch (err) {
    stats.failures++;
    debug('[alert-operator] provisioning iteration failure');
    if (err.stack) {
      debug('[alert-operator] stack: %s', err.stack.replace('\n', '\\n'));
    }
  }

  await base.testing.sleep(this.provisionIterationInterval);
} while(that.__keepRunning && !process.env.PROVISION_ONCE);
```

provisioner/provision.js
```diff
@@ -79,61 +80,90 @@ var stats = {
...
 
-    that.__watchDog.touch();
+    this.__watchDog.touch();
 
     // We should make sure that we're not just permanently failing
     // We also don't want to
-    if (stats.iterations > 20 && stats.failures > stats.success) {
-      debug('[alert-operator] killing provisioner because it has run ' +
-            'for a while but has failed lots of iterations');
+    if (stats.failure > MAX_FAILURES) {
```
jonasfj: I think we're counting errors wrong here.
If we have `MAX_FAILURES` in a row, then we should crash.

Hence, at every successful iteration we should just reset the `failures` counter. There is little point in having an absolute counter, I suspect the errors we get are from AWS.... but who knows...

provisioner/provision.js
```diff
@@ -164,7 +194,7 @@ Provisioner.prototype.runAllProvisionersOnce = function () {
   ]);
 
   p = p.then(function (res) {
-    var workerTypes = shuffle.knuthShuffle(res[0].slice(0));
+    var workerTypes = res[0];
```
jonasfj: I'm just curious, but wasn't this a good thing?

config/defaults.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     publishMetaData:      'false',
     statsComponent:       undefined,
     awsKeyPrefix:         undefined,
-    iterationInterval:    1000 * 10,
+    iterationInterval:    1000 * 1,
```
jonasfj: hmm... Maybe what we need is `minIterationInterval` as a setting for the minimum time between start of iterations.

Hence, if an iteration takes `x` ms, we sleep for `Math.max(minIterationInterval - x, 0)` ms before next iteration. So if iterations takes longer than 75s we don't sleep between them... Just an example...

provisioner/provision.js
jonasfj: Oh, just noticed my code doesn't catch errors... so if you want that semantic... you probably need to wrap with a `try {} catch(err) {process.exit()}` around the loop...

provisioner/provision.js
jhford: i always make that mistake!

provisioner/provision.js
jhford: it mostly did, but I wanted to be super-duper-dont-page-me-at-3am-sure that if it gets stuck that it really truly will die.

provisioner/provision.js
jhford: Yah, bad idea, left over from a long time ago.  I'll do that now.

provisioner/provision.js
jhford: My original plan was to do something about failure percentages and all that.  I think consecutive errors is enough

provisioner/provision.js
jonasfj: Oh, yeah I agree with that :)
I would just rely on `process.exit` to avoid having to even think about the other flow... not that it can do any harm :)

provisioner/provision.js
jhford: great idea!  i did this, though i also added in a delay to the loop using the same delayer thing i wrote to replace the spawn delays.

provisioner/provision.js
jhford: sorta... i realised it wasn't actually helping to do what I wanted to.  I might use this shuffler later on for round robining in case we decide to do random order instead of sequentially from a random order of worker names

config/defaults.js
jhford: Hmm, yes, we could do that.  We also have a hard limit of 10 minutes before we process.exit() and I'd like to let the EC2 api cool down between iterations.  If we have a >75s iteration, we probably just melted the AWS api down, and we should take a break of 75s anyway.

PR #31, opened 2015-07-08
ES6/7 fixes and roundrobin

jhford:
Rather than trying to trace through all the random issues, I decided to finish moving the important things over to es6/7so that I could be sure related to `that.` weren't causing trouble as I've seen before.  I also implemented the round robin stuff that @jonasfj has wanted for a while.

The biggest outstanding question I have is whether I should be doing:
```javascript
// assume x is a list of promises
async function () {
  return await Promise.all(x)
}

// or

async function () {
  return Promise.all(x)
}

// or

function () {
  return Promise.all(x)
}
```

### DIFF
--- a/provisioner/aws-manager.js
+++ b/provisioner/aws-manager.js
@@ -68,9 +68,7 @@ module.exports = AwsManager;
  * Update the state from the AWS API and return a promise
  * with no resolution value when completed.
  */
-AwsManager.prototype.update = function () {
-  var that = this;
-
+AwsManager.prototype.update = async function () {
   // We fetch the living instance and spot requests separate from the dead ones
   // to make things a little easier to work with as there's really very little
   // in the provisioner which requires info on dead instances and spot requests
@@ -78,12 +76,12 @@ AwsManager.prototype.update = function () {
   // The choice in which bucket each instance or request should belong in comes
   // down to whether or not the resource is awaiting or currently working or
   // needs to be tidied up after
-  var p = Promise.all([
-    that.ec2.describeInstances({
+  var res = await Promise.all([
+    this.ec2.describeInstances({
       Filters: [
         {
           Name: 'key-name',
-          Values: [that.keyPrefix + '*'],
+          Values: [this.keyPrefix + '*'],
         },
         {
           Name: 'instance-state-name',
@@ -91,22 +89,22 @@ AwsManager.prototype.update = function () {
         },
       ],
     }),
-    that.ec2.describeSpotInstanceRequests({
+    this.ec2.describeSpotInstanceRequests({
       Filters: [
         {
           Name: 'launch.key-name',
-          Values: [that.keyPrefix + '*'],
+          Values: [this.keyPrefix + '*'],
         }, {
           Name: 'state',
           Values: ['open'],
         },
       ],
     }),
-    that.ec2.describeInstances({
+    this.ec2.describeInstances({
       Filters: [
         {
           Name: 'key-name',
-          Values: [that.keyPrefix + '*'],
+          Values: [this.keyPrefix + '*'],
         },
         {
           Name: 'instance-state-name',
@@ -114,11 +112,11 @@ AwsManager.prototype.update = function () {
         },
       ],
     }),
-    that.ec2.describeSpotInstanceRequests({
+    this.ec2.describeSpotInstanceRequests({
       Filters: [
         {
           Name: 'launch.key-name',
-          Values: [that.keyPrefix + '*'],
+          Values: [this.keyPrefix + '*'],
         }, {
           Name: 'state',
           Values: ['cancelled', 'failed', 'closed', 'active'],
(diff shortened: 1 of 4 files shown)


jonasfj:
Regarding:
```js
// assume x is a list of promises
async function () {
  return await Promise.all(x)
}
// or
async function () {
  return Promise.all(x)
}
// or
function () {
  return Promise.all(x)
}
```
They are all semantically equivalent. Well, almost if you do `async function` then any error thrown synchronously by the code in your expression `x` will still result in a rejected promise. Where as this won't be the case for the last form.
Example:
```js
var f1 = async function() {
  throw new Error("Synchronous Error");
  return Promise.all(...);
};

var f2 = function() {
  throw new Error("Synchronous Error");
  return Promise.all(...);
};

f1.catch(err => console.log("Got the error"));
f2.catch(err => console.log("Got the error")); // DOES NOT WORK

// If you use f1 and f2 in an async function, you won't notice the difference
async () => {
  // THis works
  try {
    await f1();
  } catch(err) {
    console.log("Got error");
  }

  // This works too
  try {
    await f2();
  } catch(err) {
    console.log("Got error");
  }
}:
```

Generally, I favor `function() {return Promise.all(...);}` because it has less overhead, but it's probably irrelevant. The other two solutions are also okay, but `return await` rarely makes sense. It's a matter of taste, and if what you do in the async function can throw an error, you might want to consider `async function`, it's probably also a nicer way to see that it returns a promise...

Hmm.... I think I just argued against my own policy of using `function() {return Promise.all(...);}`...  I hate when I disagree with myself, now I'll probably have to start an edit war with myself...

Anyways, either one is okay.

Note, `Promise.all` can be tricky, if there is an error it'll reject before all promises are resolved!

provisioner/aws-manager.js
```diff
@@ -68,9 +68,7 @@ module.exports = AwsManager;
  * Update the state from the AWS API and return a promise
  * with no resolution value when completed.
  */
-AwsManager.prototype.update = function () {
-  var that = this;
-
+AwsManager.prototype.update = async function () {
```
jonasfj: I'm curious, but why not go all the way and convert to `class` style constructs?

provisioner/aws-manager.js
```diff
@@ -189,10 +173,9 @@ AwsManager.prototype.update = function () {
  * state about which type/region/zone combinations are not working well right
  * now
  */
-AwsManager.prototype.handleStalledRequests = function (spotReqs) {
-  var that = this;
-  return Promise.all(Object.keys(spotReqs).map(function (region) {
-    return that.killCancel(region, [], spotReqs[region].map(function (sr) {
+AwsManager.prototype.handleStalledRequests = async function (spotReqs) {
+  await Promise.all(Object.keys(spotReqs).map(region => {
```
jonasfj: Tip: lodash has a `_.map` that works well on objects... I prefer to avoid ever having to iterate on keys in objects, but convert to list of objects... But when you have to `_.map({key: value}, (value, key) => {})` works very well, only slight annoyance is that it flips keys and values in the arguments, but I can live with that, as if the object is a list, the key is the index.

provisioner/aws-manager.js
```diff
@@ -283,32 +265,24 @@ AwsManager.prototype._compareStates = function (newState, previousState, deadSta
...
-  var allInstancesInNewState = newState.instances.map(function (instance) {
-    return instance.InstanceId;
-  });
-  var allRequestsInNewState = newState.requests.map(function (request) {
-    return request.SpotInstanceRequestId;
-  });
+  let allInstancesInPreviousState = previousState.instances.map(i => i.InstanceId);
+  let allRequestsInPreviousState = previousState.requests.map(r => r.SpotInstanceRequestId);
+  let allInstancesInNewState = newState.instances.map(i => i.InstanceId);
+  let allRequestsInNewState = newState.requests.map(r => r.SpotInstanceRequestId);
```
jonasfj: nice! :)

provisioner/aws-manager.js
```diff
@@ -607,7 +577,7 @@ AwsManager.prototype.instancesInRegion = function (region) {
   if (typeof region === 'string') {
     region = [region];
   }
-  return this.__apiState.instances.filter(function (instance) {
+  return this.__apiState.instances.filter(instance => {
     return region.includes(instance.Region);
   });
```
jonasfj: If aggressive could be written as:
`return this.__apiState.instances.filter(i => region.includes(i.Region));` and still fit below 80 chars :)

provisioner/aws-manager.js
```diff
@@ -1047,27 +1000,21 @@ AwsManager.prototype.createKeyPair = function (workerName) {
...
+  for (let region of this.ec2.regions) {
+    let matchingKey = res[region].KeyPairs[0];
+    if (!matchingKey) {
+      debug('creating missing key %s in %s', keyName, region);
+      toCreate.push(this.ec2.importKeyPair.inRegion(region, {
+        KeyName: keyName,
+        PublicKeyMaterial: this.pubKey,
+      }));
+    }
+  }
```
jonasfj: In this case, `filter/map` might work nicely.
```
  await Promise.all(
    this.ec2.regions
            .filter(region => res[region].KeyPairs[0])
            .map(region => {
    debug('creating missing key %s in %s', keyName, region);
    return this.ec2.importKeyPair.inRegion(region, {
      KeyName: keyName,
      PublicKeyMaterial: this.pubKey,
    }));
  }));
```
I'm not sure how to indent `this.ec2.regions` when breaking like this... but it reads nicely...

provisioner/data.js
```diff
@@ -211,39 +211,31 @@ WorkerType.create = function (workerType, properties) {
...
-WorkerType.loadAll = function () {
-  var workers = [];
+WorkerType.loadAll = async function () {
+  let workers = [];
 
-  var p = base.Entity.scan.call(this, {}, {
+  await base.Entity.scan.call(this, {}, {
     handler: function (item) {
       workers.push(item);
     },
```
jonasfj: You'll enjoy `{handler(item) { workers.push(item); }}` or `{handler: item => workers.push(item)}` even more :)

First one uses method in dictionary syntax...

provisioner/data.js
```diff
@@ -261,9 +253,8 @@ WorkerType.prototype.json = function () {
  * and optionally a single property from it.
  */
 WorkerType.prototype.getInstanceType = function (instanceType) {
-  var types = this.instanceTypes.filter(function (t) {
-    return t.instanceType === instanceType;
-  });
+  let types = this.instanceTypes.filter(t => t.instanceType === instanceType);
```
jonasfj: Tip: lodash has `_.filter(this.instanceTypes, {instanceType: instanceType})` which can also be written as:
`_.filter(this.instanceTypes, {instanceType})` because key and variable name for value is the same.

Just a tip, I haven't found many places to use this yet...

jonasfj:
Got through the ES6 rewrite parts that looks really nice! :)
I see you're better at embracing `for of` than me, I still use `forEach` too much...

provisioner/provision.js
```diff
@@ -192,9 +194,50 @@ Provisioner.prototype.runAllProvisionersOnce = async function () {
...
+      debug('%s needs %d capacity destroyed', worker.workerType, capToKill);
+      await this.awsManager.killCapacityOfWorkerType(
+            worker, capToKill, ['pending', 'spotReq']);
+    } else {
+      debug('%s needs no changes', worker.workerType);
+    }
+  }
+
+  var d = delayer(500);
+  var longD = delayer(2000);
```
jonasfj: This is nice... you should make these constants... in global scope like:
```js
const DELAY_BETWEEN_SPAWNING = delayer(500);

// later down in async function
  ....
  await DELAY_BETWEEN_SPAWNING();
```

jonasfj:
@jhford, so round robin here is only which order we provisioner for workerTypes right?

It's not round robin between 10% of the cheapest region/instanceType configurations, right?

provisioner/provision.js
```diff
@@ -192,9 +194,50 @@ Provisioner.prototype.runAllProvisionersOnce = async function () {
...
+
+  // We want to have a maximum number of attempts
+  var attemptsLeft = forSpawning.length * 2;
```
jonasfj: I'm not sure it's a good idea to use attempts across all workerTypes... How about attempts was a property on the `toSpawn` object? Then default to 3 for all workerTypes...

Hmm, maybe, maybe not.... WorkerTypes aren't dependent on single region, so no reason one should fail more than the others...

jonasfj:
@jhford, comments are suggestions I think, all of them are, so `r+`...

provisioner/provision.js
jhford: so what i'm doing is getting a list of things to spawn, ordered by workerType.  I'm then shuffling them up and so I no longer have them separated by workerType.  I would have to maintain a retry count per workerTYpe, which isn't important.  Instead, I decided to just say that you have 2x the total number of all things to spawn attempts before failing.

Not perfect, but we're going to retry in 75s anyway...

jonasfj: Yeah... maybe we don't even need retries... :)

PR #32, opened 2015-07-13
Cleanups and state improvements

jhford:
A whole bunch of cleanups, and also the first stage in storing the state of each worker type in a list.  I really would prefer to maintain the structure of the AWS API returns, but when we have a limit for how large an entity can be, I didn't want to waste space.  I can't quite remember how large the JSONType can be, what do you think about storing it in this way?  I'm not sure if we can break it down further in a meaningful way, so I hope that Entities can handle this.

For the time being, I'm leaving the old aws state in there because i need to update the tools site as well.  The plan is deploy this then update the tools ui to consume this then update the the server to delete the legacy code around having an aws manager in the api server.

### DIFF
--- a/lib/aws-manager.js
+++ b/lib/aws-manager.js
@@ -3,7 +3,7 @@
 let Promise = require('promise');
 let assert = require('assert');
 let debug = require('debug')('aws-provisioner:aws-manager');
-let objFilter = require('../lib/objFilter');
+//let objFilter = require('../lib/objFilter');
 let shuffle = require('knuth-shuffle');
 let taskcluster = require('taskcluster-client');
 let series = require('./influx-series');
@@ -196,7 +196,16 @@ AwsManager.prototype.update = async function () {
   let stateDifferences = this._compareStates(this.__apiState, this.__previousApiState, this.__deadState);
   this._reconcileStateDifferences(stateDifferences, this.__deadState, this.__apiState);
 
-  await this.handleStalledRequests(filteredSpotRequests.stalled);
+  try {
+    await this.handleStalledRequests(filteredSpotRequests.stalled);
+  } catch (err) {
+    debug('error handling stalled spot requests');
+    debug(err);
+    if (err.stack) {
+      debug(err.stack);
+    }
+    throw err;
+  }
 
   // We want to make sure that our internal state is always up to date when
   // we fetch the updated state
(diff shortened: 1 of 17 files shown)


jonasfj:
You can store atmost 1MiB across all properties of an entity. The `base.Entity` does store a version and for things like JSONType it stores an interview extra... It doesn't add up to much more than a few bytes, but you have less than 1 MiB for all properties combined!

The alternative is to use blob storage, but my guess is this will likely work... If you try to minimize what datat you store.

A safe estimate would be 100 bytes per instance or spot request, then 900kb gives you 9000 instances or spot request, right? I think that'll do for a while....oh and catch exceptions from saving this and alert on them, but don't crash!

jhford:
@jonasfj i'd be up for trying out blob storage... do we already have wrappers for that or would I need to figure that out?

I think the max we have is 400-500 instances/requests... this should do for the time being

config/development.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     id:                   'aws-provisioner2-dev',
     workerTypeTableName:  'AwsWorkerTypesDev2',
     secretTableName:      'ProvisionerSecretsDev',
+    workerStateTableName: 'ProvisionerWorkerStateDev1',
```
gregarndt: why Dev1?

jhford: because i was trying to figure out an issue and screwed up the non-1 table... it's not as easy to delete tables as it is to make them.

lib/aws-manager.js
```diff
@@ -123,7 +128,36 @@ AwsManager.prototype.update = async function () {
...
+  let res;
+  try {
+    res = await Promise.all(statusPromises);
+  } catch (err) {
+    debug('Error running status and pricing promises');
+    debug(err);
+    if (err.stack) {
+      debug(err.stack);
+    }
+    throw err;
```
gregarndt: Does this end up crashing whatever process called update()?

jonasfj:
@jhford,
> I think the max we have is 400-500 instances/requests... this should do for the time being
"now the time being" is a bit short sighted, let's get it done and then be done with it :)

Note, we don't have a way for auth.tc.net to grant you access to a container... We should make this...
But since you already put azure credentials in config... right?
As far as I can see you never fixed it: https://github.com/taskcluster/aws-provisioner/blob/eb31d0dba97e3ef42ada5e83a86fb4e804c5599f/bin/server.js#L98
(I note, that I pointed delegation from auth out as the preferred azure access option a few times)

Anyways, since we have azure credentials, you can just use those with `azure-storage` library. For a tested more promisifed wrapper just steal:
https://github.com/taskcluster/taskcluster-queue/blob/50dab93f468d18235033f0e94dfffb4b7feef80d/queue/blobstore.js

That should make it easy to do...

jonasfj:
I don't see anything bad... lots of try/catch/re-throw that might not be needed... but I suppose we're getting desperate to find out where it hangs :)

Anyways, I didn't make it all the way through this... @jhford, I would recommend moving to blob storage, and then well, ask for a quick review... or just merge it, honestly at this size and the amount of things merged in, you might be better off deploying it while waiting for review :)

jhford:
Ok! because of how thing are stored (no longer matching ec2 api names) i have a small patch to tools that I've verified to work.  I'll push that along with this code.

jhford:
landed https://github.com/taskcluster/taskcluster-tools/commit/29a777779021f7abd3cfe319f6fd0ec45da2cf4d to deal with the changes to state format

PR #33, opened 2015-07-13
update secret docs

djmitche:
I'll be happy to rebase this after #32 lands.


jhford:
Looks good to me, other than the double quotes.  The provisioner uses single quotes everywhere and instead escapes where needed.  I have eslint as the first step of the `npm test` steps, so if you get mocha unit tests running, you can assume that any linter output was warnings only

djmitche:
I get 46 problems from `npm test` on master, and now 46 on my branch (just force pushed)

jhford:
46 unit test failures or 46 linting issues?

djmitche:
✖ 46 problems (25 errors, 21 warnings)

so all linting failures

jonasfj:
Looks good to me... Only note is that instead of saying "token" it would be nice to use the property name used in user-data. And then use it in back-ticks so it's clear that it's a property. Notice docs. etc. can contain markdown.

djmitche:
updated to name the `securityToken` property.

jhford:
Dustin, if you rebase onto current master, I can happily merge when it's a good time to restart the provisioner

djmitche:
rebased

PR #34, opened 2015-07-16
check that amis exist

jhford:
@gregarndt might like this... we should check that given AMIs exist before creating the worker type!

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -207,7 +207,8 @@ api.declare({
   // We want to make sure that all AMIs that we are submitting are valid
   let missing = [];
   await Promise.all(input.regions.map(async (def) => {
-    if (await amiExists(this.ec2, def.launchSpec.ImageId, def.region)) {
+    let exists = await amiExists(this.ec2[def.region], def.launchSpec.ImageId);
+    if (!exists) {
       missing.push({imageId: def.launchSpec.ImageId, region: def.region});
     }
   }));
@@ -320,11 +321,13 @@ api.declare({
   // We want to make sure that all AMIs that we are submitting are valid
   let missing = [];
   await Promise.all(input.regions.map(async (def) => {
-    if (await amiExists(this.ec2, def.launchSpec.ImageId, def.region)) {
+    let exists = await amiExists(this.ec2[def.region], def.launchSpec.ImageId);
+    if (!exists) {
       missing.push({imageId: def.launchSpec.ImageId, region: def.region});
     }
   }));
   if (missing.length > 0) {
+    debug(missing);
     return res.status(400).json({
       message: 'ami does not exist',
       missing: missing,
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -1,6 +1,7 @@
 let debug = require('debug')('routes:v1');
 let base = require('taskcluster-base');
 let taskcluster = require('taskcluster-client');
+let amiExists = require('../check-for-ami');
 let _ = require('lodash');
 let rp = require('request-promise');
 let url = require('url');
@@ -202,6 +203,22 @@ api.declare({
   // TODO: If workerType launchSpecification specifies scopes that should be given
   //       to the workers using temporary credentials, then you should validate
   //       that the caller has this scopes to avoid scope elevation.
+
+  // We want to make sure that all AMIs that we are submitting are valid
+  for (let region of input.regions) {
+    let validAmi = await amiExists(this.ec2, region.launchSpec.ImageId, region.region);
+    if (!validAmi) {
+      res.status(400).json({
+        message: 'ami does not exist',
+        ami: region.launchSpec.ImageId,
+        region: region.region,
+      });
+      return;
+    }
+  }
+
+  // We want to make sure that every single possible generated LaunchSpec
+  // would be valid before we even try to store it
   try {
     await validateWorkerType(this, workerType, input);
   } catch (err) {
@@ -298,6 +315,18 @@ api.declare({
   let workerType = req.params.workerType;
 
   let modDate = new Date();
+
+  // We want to make sure that all AMIs that we are submitting are valid
+  for (let region of input.regions) {
+    let validAmi = await amiExists(this.ec2, region.launchSpec.ImageId, region.region);
+    if (!validAmi) {
+      return res.status(400).json({
+        message: 'ami does not exist',
+        ami: region.launchSpec.ImageId,
+        region: region.region,
+      });
+    }
+  }
 
   input.lastModified = modDate;
 
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -1,7 +1,7 @@
 let debug = require('debug')('routes:v1');
 let base = require('taskcluster-base');
 let taskcluster = require('taskcluster-client');
-let amiExists = require('../check-for-ami');
+let amiExists = require('./check-for-ami');
 let _ = require('lodash');
 let rp = require('request-promise');
 let url = require('url');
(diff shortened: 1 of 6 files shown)


lib/check-for-ami.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (result.Images.length !== 1) {
+    throw new Error('found more images than expected');
+  }
+
+  if (result.Images[0].ImageId === ami) {
+    return true;
+  } else {
+    throw new Error('something wacky happened');
```
gregarndt: Love the humor here, but if looking at logs, we might want to include something about what "wacky" is :)

lib/check-for-ami.js
```diff
@@ -0,0 +1,46 @@
...
+      assert(!region, 'this ec2 module doesnt understand region');
+      result = await ec2.describeImages(request);
+    }
+    debug('result from ec2: %j', result);
+  } catch (err) {
+    debug('did not find %s', ami);
+    return false;
+  }
+
+  if (result.Images.length !== 1) {
```
gregarndt: so, when searching for a particular ami, more than one could return?

lib/routes/v1.js
```diff
@@ -169,6 +183,18 @@ api.declare({
 
   let modDate = new Date();
 
+  // We want to make sure that all AMIs that we are submitting are valid
+  for (let region of input.regions) {
+    let validAmi = await amiExists(this.ec2, region.launchSpec.ImageId, region.region);
+    if (!validAmi) {
```
gregarndt: I'm not aware of the behavior, but what happens when the underlying code throws an error? does our api handling return back a certain status code then with the error?  Do we want to catch the error and then include the ami and region that was used in the response?

jonasfj: API wrapper assumes handlers returns a promise... and if the promise is rejected, it'll return 500 with an incidentid, error message (which may be sensitive) is printed to log along with the incidentid.

lib/check-for-ami.js
jhford: yah... true.

lib/check-for-ami.js
jhford: With the ec2 api, i'm not going to rule out anything.  This is purely defensive programming.

gregarndt:
:+1:

jhford:
@jonasfj this might have bitrotted a bit, but what do you think about this check?

lib/check-for-ami.js
```diff
@@ -0,0 +1,46 @@
...
+      'all',
+    ],*/
+    ImageIds: [ami],
+  };
+
+  let result;
+  try {
+    // If we're using the multi-region wrapper, we handle things
+    // slightly differently
+    if (ec2.describeImages && ec2.describeImages.inRegion) {
```
jonasfj: Do we have a need for both? can't we just assert that you always give `region` and an instance of `multi-region wrapper` ?

lib/check-for-ami.js
```diff
@@ -0,0 +1,46 @@
...
+      result = await ec2.describeImages.inRegion(region, request);
+    } else {
+      assert(!region, 'this ec2 module doesnt understand region');
+      result = await ec2.describeImages(request);
+    }
+    debug('result from ec2: %j', result);
+  } catch (err) {
+    debug('did not find %s', ami);
+    return false;
```
jonasfj: Does it return an empty list if there is no matches or does it throw an error.
If it returns an empty list we should not handle the error, as it indicate inability to check for existence, right?

lib/routes/v1.js
```diff
@@ -169,6 +183,18 @@ api.declare({
 
   let modDate = new Date();
 
+  // We want to make sure that all AMIs that we are submitting are valid
+  for (let region of input.regions) {
+    let validAmi = await amiExists(this.ec2, region.launchSpec.ImageId, region.region);
```
jonasfj: I would do:
```js
let missing = [];
await Promise.all(input.regions.map(async (def) => {
  if (await amiExists(this.ec2, def.launchSpec.ImageId, def.region)) {
    missing.push({imageId: def.launchSpec.ImageId, region: def.region});
  }
}));
if (missing.length > 0) {
   .... // Send error
}
```
or even shorter:
```js
let missingAMI = _.every(await Promise.all(input.regions.map(def => {
  return amiExists(this.ec2, def.launchSpec.ImageId, def.region);
})));
if (missingAMI) {
   .... // Send error without saying which AMI is missing...
}
```

Either way, using `Promise.all` makes sense as we get the requests in parallel.. As the requests goes out to different regions, there is no reason not to make them in parallel.

jonasfj:
Minor suggestions, generally I think it's nice to do this..
But we still have to survive cases where someone deletes an AMI after creating the workerType.

jonasfj:
`r+`, but please do consider some of the suggestions...

lib/check-for-ami.js
jhford:
```javascript
> a = {}
{}
> if (a.b.c) {
... console.log('hi');
... }
TypeError: Cannot read property 'c' of undefined
    at repl:1:8
    at REPLServer.defaultEval (repl.js:132:27)
    at bound (domain.js:254:14)
    at REPLServer.runBound [as eval] (domain.js:267:12)
    at REPLServer.<anonymous> (repl.js:279:12)
    at REPLServer.emit (events.js:107:17)
    at REPLServer.Interface._onLine (readline.js:214:10)
    at REPLServer.Interface._line (readline.js:553:8)
    at REPLServer.Interface._ttyWrite (readline.js:830:14)
    at ReadStream.onkeypress (readline.js:109:10)
```

I just like do be extra careful when checking.

lib/check-for-ami.js
jhford: It throws an error, sadly.  There might be some property on the error object that will tell us this, but the outcome for both failing to check and not having the image is the same: we don't process the request.

lib/routes/v1.js
jhford: This was written when we were thinking that Promise.all might be related to provisioner freezing and I was avoiding using it.  Yes, we can do it with Promise.all!

jhford:
The problem we currently don't have a way to bubble up errors from the provisioner background process to be visible to a consumer without dropping down to Heroku logs.

jhford:
This should still be done, but someone will need to unbitrot this patch first.  I'll see if I can do it this week.

jhford:
@jonasfj did you want another look?  if not, I'll be pushing this tomorrow

src/check-for-ami.js
```diff
@@ -0,0 +1,45 @@
...
+    ],*/
+    ImageIds: [ami],
+  };
+
+  let result;
+  try {
+    // If we're using the multi-region wrapper, we handle things
+    // slightly differently
+    if (ec2.describeImages && ec2.describeImages.inRegion) {
+      result = await ec2.describeImages.inRegion(region, request);
```
jonasfj: didn't we drop multi-region-aws-thingy?

jonasfj: Ì think you forgot to update this patch...

src/api-v1.js
```diff
@@ -299,6 +317,20 @@ api.declare({
...
 
+  // We want to make sure that all AMIs that we are submitting are valid
+  let missing = [];
+  await Promise.all(input.regions.map(async (def) => {
+    if (await amiExists(this.ec2, def.launchSpec.ImageId, def.region)) {
+      missing.push({imageId: def.launchSpec.ImageId, region: def.region});
+    }
+  }));
+  if (missing.length > 0) {
+    return res.status(400).json({
```
jonasfj: nit, if you are on a recent tc-lib-api, there is a reportError function which will do great here...

jonasfj:
@jhford, I think there is some bitrot here... just tiny things..

src/check-for-ami.js
jhford: i wrote this patch knowing the drop would happen, but I'll remove the transitional code any way!

PR #36, opened 2015-07-22
load worker type entities individually instead of in a scan

jhford:
I'm wondering if there's something happening in the scan operation while building the list of worker entities.  This changes things to make it load the list of strings, which are much simpler, then load each workerType just before it is about to have its bids determined.

There is a loss of atomicity in the load operations here now.  There are two cases which we care about: creating and deleting a worker type.  The time delta between loading the list of names and loading the definitions is only important for these operations and only for the time period between the two operations.  The majority of code between listing the names and loading the definitions is not too large, just the time to lead the remaining entities.  There are no ec2 api calls until all workerType definitions are loaded.

When creating a workerType between the list() and load() call, we simply ignore that worker type for that given iteration.  The iteration is marked as success and the next iteration is the first one with the new worker type.

When deleting a workerType between the list() and load() call, we will try to load the workerType which will fail.  The ResourceNotFound exception will cause that workerType to fail to run its provisioning logic and will prematurely end that iteration, marking it as failure.  This will not kill the provisioner as we need >10? consecutive failing iterations to kill things.  The next iteration will be able to complete as designed.

This is more a diagnostic than a desired change... I'm just trying to reduce complexity where I can.  Building a list of strings is less complex that building a list of entities.

### DIFF
--- a/lib/provision.js
+++ b/lib/provision.js
@@ -158,10 +158,10 @@ Provisioner.prototype.stop = function () {
  * Run provisioners for all known worker types once
  */
 Provisioner.prototype.runAllProvisionersOnce = async function () {
-  let workerTypes;
+  let workerNames;
   try {
-    workerTypes = await this.WorkerType.loadAll();
-    debug('loaded worker types');
+    workerNames = await this.WorkerType.listWorkerTypes();
+    debug('loaded worker type names');
   } catch (err) {
     debug('error loading worker types');
     debug(err);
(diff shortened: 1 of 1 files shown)


jonasfj:
This won't diagnose anything... the objects are still created in `listWorkerTypes`, which is implement using:
```js
    await base.Entity.scan.call(this, {}, {
      handler: function (item) {
        names.push(item.workerType);
      },
    });
```
The only difference from `loadAll` is `item.workerType` instead of just `item`...There is no way this works better... unless miss a subtle implementation in `worker-type.js`, but then it's not `base.Entity`.

jhford:
this wasn't needed and we're no longer freezing.  no longer want this.

PR #37, opened 2015-07-23
Remove loose option

jonasfj:
We probably should do babel loose translation...

I have no idea if this helps... but it can't hurt when facing what we're currently dealing with.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,3 @@
 {
-  "stage":  1,
-  "loose":  "all"
+  "stage":  1
 }


jonasfj:
@jhford, actually everywhere else that I use `babel-node` I also install `kexec` it's used by `babel-node` if available:
https://github.com/babel/babel/blob/b9dde49ab6c3c86674713894b6c2eec665e8c1c5/packages/babel-cli/src/babel-node.js#L433-L434

So maybe the subprocess is somehow dead without the `babel-node` process detecting it and exiting...
that would explain all the symptoms we are seeing... As it appears to me that it's just like it stops working completely... Ie. timeouts and everything...

Note, I've seen the error twice in logs and it's about the same...

jonasfj:
My bad.. it's only once that we've seen the hang around `base.Entity.scan´ not multiple times.
So this could be completely unrelated...

jhford:
Ok, done both of these things with c0f670672f9d880092e6d31dc43e10aa819f502c

PR #38, opened 2015-07-24
Bug 1184272: specify all API scopes as [[str]]

djmitche:
I think this is the last one!


jhford:
As long as the linter and unit tests run clean, this looks good to me!

djmitche:
Do those not run in travis?

jhford:
it looks like travis is not running on this commit, or at least I can't see it doing that

jhford:
Linter and tests pass when I run the tests manually.  Merging.

PR #41, opened 2015-08-05
Ensure that users can only create workers with scopes that they have

jhford:
Including unit tests!

### DIFF
--- a/lib/routes/v1.js
+++ b/lib/routes/v1.js
@@ -105,6 +105,15 @@ api.declare({
     return;
   }
 
+  // Figure out which scopes the client needs
+  let requiredScopes = this.WorkerType.scopes(input);
+  // Now check for them if there are any
+  if (requiredScopes.length > 0) {
+    if (!req.satisfies(requiredScopes)) {
+      return;
+    }
+  }
+
   // TODO: If workerType launchSpecification specifies scopes that should be given
   //       to the workers using temporary credentials, then you should validate
   //       that the caller has this scopes to avoid scope elevation.
@@ -218,6 +227,15 @@ api.declare({
   input.lastModified = modDate;
 
   if (!req.satisfies({workerType: workerType})) { return undefined; }
+
+  // Figure out which scopes the client needs
+  let requiredScopes = this.WorkerType.scopes(input);
+  // Now check for them if there are any
+  if (requiredScopes.length > 0) {
+    if (!req.satisfies(requiredScopes)) {
+      return;
+    }
+  }
 
   try {
     this.WorkerType.testLaunchSpecs(input, this.keyPrefix, this.provisionerId, this.provisionerBaseUrl);
--- a/lib/worker-type.js
+++ b/lib/worker-type.js
@@ -339,6 +339,59 @@ WorkerType.prototype.testLaunchSpecs = function () {
 };
 
 /**
+ * Return a list of every scope for every region and instance type combination.
+ * Currently, scope [1,2] is considered unique compared to [2,1].  it would be
+ * nice if we could convert the array of arrays into a set of sets for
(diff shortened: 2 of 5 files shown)


jhford:
@jonasfj can you do a pretty good check on the scopes method that I added, i want to be extra certain that I didn't goof on the logic!

lib/worker-type.js
```diff
@@ -339,6 +339,59 @@ WorkerType.prototype.testLaunchSpecs = function () {
...
+ * nice if we could convert the array of arrays into a set of sets for
+ * comparisons
+ */
+WorkerType.scopes = function (worker) {
+  // Generate the complete list of scopes;
+  let scopes = [];
+
+  // Take an object with a .scopes property and list add all of its (mostly)
+  // unique scopes to the list of scopes to return
+  var y = function(x) {
```
jhford: ugh, that should be a let not a var

lib/routes/v1.js
```diff
@@ -105,6 +105,15 @@ api.declare({
     return;
   }
 
+  // Figure out which scopes the client needs
+  let requiredScopes = this.WorkerType.scopes(input);
+  // Now check for them if there are any
+  if (requiredScopes.length > 0) {
+    if (!req.satisfies(requiredScopes)) {
```
jonasfj: I think this has to be `req.satisfies([requiredScopes])`..

The argument given to `satisfies` is a set of scope-sets in disjunctive form.
e.g. `req.satisfies([[a], [b,c]])` means "`a` or (`b` and `c`)"

lib/worker-type.js
```diff
@@ -339,6 +339,59 @@ WorkerType.prototype.testLaunchSpecs = function () {
...
+    y(instanceType);
+  }
+
+  return scopes; 
+};
+
+/**
+ * Make the scopes method available on instances as well
+ */
+WorkerType.prototype.scopes = function () {
```
jonasfj: This overwrites a property... So probably a bad idea...

lib/routes/v1.js
jhford: if I'm remembering the code I wrote properly, `requiredScopes` will have that form.

From the later test:
```javascript
+      subject.scopes(wType).should.eql([
+        ['gScope', 'g'],
+        ['rScope1', 'r'],
+        ['rScope2'],
+        ['r'],
+        ['tScope1', 't'],
+        ['tScope2'],
+        ['t'],
+      ]);
```

lib/worker-type.js
jhford: omg.  yes.  oops.

lib/worker-type.js
```diff
@@ -339,6 +339,59 @@ WorkerType.prototype.testLaunchSpecs = function () {
...
+    y(region);
+  }
+
+  // Instance Type specific scopes
+  for (let instanceType of worker.instanceTypes) {
+    y(instanceType);
+  }
+
+  return scopes; 
+};
```
jonasfj: This seems way to complicated... and wrong.
Just do:
```js
return _.union(
  worker.scopes,
  _.flatten(worker.instanceTypes.map(t => t.scopes)),
  _.flatten(worker.regions.map(r => r.scopes))
]));
```
I assume you're code always creates the `scopes` property as an array, even if it's empty.
If you don't do that... then I recommend that you start doing so, and do a version upgrade then you can fix the old instances in the `migrate` function.

Sure you could add `|| []` everywhere but that is bad... Defaults should exists in JSON schemas and if not there, then when objects are created. But I'm pretty sure you can settle with just having them in your JSON schemas (for the input).

jonasfj:
@jhford, has some variant of this been merged?
If not the comments seems easy...

PR #42, opened 2015-09-04
Upgrade to use remote signature validation

jonasfj:
Same as others...


jonasfj:
deployed and working...

PR #44, opened 2015-09-17
Price biasing

jhford:
Can you take a good look at the lastUpdate/maxDate logic to make sure that I haven't goofed it up.  I know that I'm getting biasing in testing, so I'm pretty confident in it.

Switching to es6 classes and removing extraneous try/catches is something that I plan on doing once I have biasing in production, but as discussed it's not a priority right now.

### DIFF
--- a/lib/biaser.js
+++ b/lib/biaser.js
@@ -42,7 +42,13 @@ function Biaser (cfg) {
   assert(typeof cfg.emptyComboBias === 'number');
   this.__emptyComboBias = cfg.emptyComboBias;
 
-  // Store an empty bias table
+  // Store an empty bias table.  A bias table is a nested object which is
+  // organized by the keys Region, Zone and Type in that order.  Those three
+  // keys make up something we call a combo here.  Each combo in the table has
+  // an object which may have the kills and spawns properties.  Those
+  // properties are the number of kills for that combo and the number of spawns
+  // for that combo, respectively.  When a kills or spawns key is not defined,
+  // a value of zero should be assumed
   this.__biasTable = {};
 
   // When we set this time to be so old, it will ensure that we short circuit
--- a/lib/biaser.js
+++ b/lib/biaser.js
@@ -70,7 +70,7 @@ module.exports = Biaser;
  * assume that we no longer have valid biasing information and return a bias
  * of one.
  */
-Biaser.prototype.getBias = function(region, zone, type) {
+Biaser.prototype.getBias = function (region, zone, type) {
   assert(region);
   assert(zone);
   assert(type);
@@ -122,7 +122,7 @@ Biaser.prototype.getBias = function(region, zone, type) {
     // If we are able to calculate what the bias is, we want to weigh it a
     // little heavier so that we avoid that combo and
     let bias = 1 + this.__killRateMultiplier * killrate;
-    return bias
+    return bias;
   } else {
     // Like if we have zero spawns and zero kills, we want to give this combo
     // a slight preference
@@ -177,7 +177,7 @@ Biaser.prototype.fetchBiasInfo = async function(azinfo, types) {
     if (!biasTable[region][zone][type]) {
       biasTable[region][zone][type] = {
         kills: 0,
-        spawns: 0
+        spawns: 0,
       };
     }
     biasTable[region][zone][type].kills = killcount;
@@ -203,7 +203,7 @@ Biaser.prototype.fetchBiasInfo = async function(azinfo, types) {
     if (!biasTable[region][zone][type]) {
       biasTable[region][zone][type] = {
         kills: 0,
-        spawns: 0
+        spawns: 0,
       };
     }
     biasTable[region][zone][type].spawns = spawncount;
--- a/lib/biaser.js
+++ b/lib/biaser.js
@@ -151,7 +151,12 @@ Biaser.prototype.fetchBiasInfo = async function(azinfo, types) {
     'group by az, region, instanceType',
   ].join(' ');
 
-  let results = await this.influx.query(spotKillQuery + ';' + spotFulfilledQuery);
+  let fullQuery = spotKillQuery + ' ; ' + spotFulfilledQuery;
+  debug(fullQuery);
+
+  let results = await this.influx.query(fullQuery);
+  assert(results);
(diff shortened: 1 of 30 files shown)


jhford:
@jonasfj it might be easier to go through the review in person while we're in the same office.

lib/biaser.js
```diff
@@ -0,0 +1,214 @@
+'use strict';
```
jonasfj: not needed with babel

jonasfj: also, please write a `class`... It's just good practice :)

lib/biaser.js
```diff
@@ -0,0 +1,217 @@
...
+  this.__killRateMultiplier = cfg.killRateMultiplier;
+
+  // We want to give preference to regions where we haven't spawned anything or
+  // had any kills as a way to distribute load
+  assert(cfg.emptyComboBias);
+  assert(typeof cfg.emptyComboBias === 'number');
+  this.__emptyComboBias = cfg.emptyComboBias;
+
+  // Store an empty bias table
+  this.__biasTable = {};
```
jonasfj: You should document the table layout... ideally in a comment here.

Note: I think the current design is particularly bad. Instead of `this.__biasTable[region][zone][type]` do something  like `this.__biasTable[region+ '/' + zone + '/' + type]` and you won't have to deal with missing intermediate objects.
But most likely this isn't called very often, so you would probably be fine with an array where `region`, `zone` and `type` are properties on elements of the array.

Nested objects is complicated to code with... and invites too many bugs, if statements and nested loops...

jhford: sure, can do.  about the design, do you mean just the nested objects or do you mean overall design?

lib/biaser.js
```diff
@@ -0,0 +1,217 @@
...
+  for (let spawnpt of spawns.points) {
+    let spawncount = spawnpt[1];
+    let zone = spawnpt[2];
+    let region = spawnpt[3];
+    let type = spawnpt[4];
+    if (!biasTable[region]) {
+      biasTable[region] = {};
+    }
+    if (!biasTable[region][zone]) {
+      biasTable[region][zone] = {};
```
jonasfj: Good example of something that would be a lot simpler, if you didn't have nested objects...

Using `this.__biasTable[region+ '/' + zone + '/' + type]` is much easier... fewer corner cases...
If you must use nested objects to avoid issues with `/` potentially being used in a region, zone or type name please consider using lodash `_.get` and `_.set`, e.g. `_.set({}, ["a", "b", "c"], 4)`.

IMO, if you're scared that `/` might be used in a region, zone or type name, it might be better to just make a utility function that concatenates these into a string for you and escapes the string names... or something like that.

lib/worker-type.js
```diff
@@ -719,7 +719,10 @@ WorkerType.prototype.determineSpotBids = function (managedRegions, pricing, chan
         for (let zone of zones) {
           try {
             let potentialBid = pricingData[region][type][zone];
-            let potentialPrice = potentialBid / uf[type];
+            let bias = biaser.getBias(region, zone, type);
+            let potentialPrice = potentialBid / uf[type] * bias;
```
jonasfj: nit, and this might just be me who is bad at math... but should we maybe add a parenthesis, so it's clear what order operations are applied...

jonasfj:
generally, I think this patch is sane... but I think the killRate is a bad measure... Nodes often lives for a long time...

Perhaps duration of nodes that were killed... or number of spot requests that weren't fulfilled...  For now I like biasing regions we aren't using... 

hmm... I'm not sure what a good strategy is here... Maybe this is too complicated... we can try and roll this out, but it might be a good idea to monitor cost efficiency measures as we roll it out... These measures aren't hard to make, take a look at the PR edgar did for tc-tools and the series from docker-worker garndt can't point you at them...

jonasfj:
@jhford, r+, but please do consider refactoring the biastable thing to not used nested objects...

lib/biaser.js
jonasfj: just the nested objects :)

Otherwise the idea of creating some bias number seems great... If we based that bias on the right measures I'm a lot less sure of...

lib/biaser.js
jhford: It's just an old habbit from working on releng buildbot configs :)

jhford:
Duration of node life might have issues where each node is expected to live a different amount of time.  The great thing about this is that we can explore different options by editing that one function in one class. 

The failure case of this optimization is that we pick the wrong region/zone/type combo, not that we don't submit a request at all.

Thanks for doing this review, I'll address concerns and land tomorrow.

lib/biaser.js
jhford: Thinking over again, I really think that nested objects works well in this case.  It makes lookup a lot easier.  I'll use lodash's set and get.

jonasfj: that's okay... as long as we use `_.get/_.set` so we don't have the stupid code creating an empty object if it an entry is missing...

jhford: one day, i'll learn to love lodash and read through all it offers.

lib/worker-type.js
jhford:
```
> 2 * 3 / 4
1.5
> (2 * 3) / 4
1.5
> (2 / 3) * 4
2.6666666666666665
> 2 / 3 * 4
2.6666666666666665
```
The linter will complain if you leave in unneeded brackets saying that they're useless, and they are, so I think there's not really much value in grouping them unnecessarily.

jhford:
Oh, and we are only counting non-TC initiated kills, i.e. Amazon killing us.

lib/biaser.js
jonasfj: @jhford,
I feel I must say this... but for the record, I don't like lodash, I think features it had ought to be part of JS...  (and in fact many are with es5 and es6).

But whenever I see code that could be smarter, I usually dig into lodash to see if perhaps not it has a smart way of doing things.

jonasfj:
> Oh, and we are only counting non-TC initiated kills, i.e. Amazon killing us.

That's why I suggested looking at duration... hmm... either way we have to experiment with this...
Before we can do that, we need cost efficiency measures and pending time measures...

PR #45, opened 2015-09-21
Es6 classes

jhford:
Not super high priority, but this patch changes all non-base.Entity classes to use the es6 format.  I don't change the base.Entity based constructors because we create them programatically and I'm not sure if we can/should use the class keyword for them.

I realised when checking the patch that I should've done one patch that just did the indentation for each file in completion then the changes to the file in a second changeset for easier review.... if that's really important I'll just do this again, it's ~3 hours of work.


jonasfj:
r+... I gave it a quick look through, not detailed... so just monitor it as you roll it out :)
It's not worth rewriting history...

Note, address issues with biaser before landing it...

PR #46, opened 2015-09-30
Don't throw errors when price is high

jonasfj:
I think this is sane, feedback please...


jonasfj:
@jhford, I think this is safe to merge... so I'll try it, since it auto-deploys... worst case you can roll it back...

PR #47, opened 2015-09-30
return existing bids instead of an empty list

jhford:
Instead of returning no bids, we know that at this stage that the bids already in spotBids are OK since we didn't hit either of the max price checks.  Instead of ignoring those bids, let's submit them and let the next iteration handle the bids which couldn't get created.

I think this is slightly better behaviour, but I still appreciate you jumping in last night to get a fix out there :)

### DIFF
--- a/lib/worker-type.js
+++ b/lib/worker-type.js
@@ -794,7 +794,7 @@ WorkerType.prototype.determineSpotBids = function (managedRegions, pricing, chan
     } else {
       debug('WorkerType %s is exceeding its max price of %d with %d',
             this.workerType, this.maxPrice, cheapestPrice);
-      return [];
+      return spotBids;
     }
 
     // This is a sanity check to prevent a screw up where we theoretically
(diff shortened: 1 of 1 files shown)


lib/worker-type.js
```diff
@@ -794,7 +794,7 @@ WorkerType.prototype.determineSpotBids = function (managedRegions, pricing, chan
     } else {
       debug('WorkerType %s is exceeding its max price of %d with %d',
             this.workerType, this.maxPrice, cheapestPrice);
-      return [];
+      return spotBids;
```
jonasfj: Why is this inside a loop... I'm not sure... ie. how can `cheapestPrice` be different in each iteration of the `while (change > 0) {` loop???

I might be missing something... But why don't we refactor into some thing sane...

1) determine list of bids for various machine/region/zone/capacity combinations
2) `while(change > 0) {` pick random bids to make from 3 cheapest options below `maxPrice`, if there is nothing below `maxPrice` pick randomly between all options give `price = maxPrice` (so we'll get nodes when capacity is available).

Not sure about the last part... But definitely simplify this code...

jonasfj:
`r+`, please land this... The comment is unrelated to this PR, just a general issue that really ought to be considered. Much of the code here is doing simple things is very complex ways.

PR #48, opened 2015-09-30
Static transpile of provisioner

jhford:
use babel in the package.json file instead of doing everything dynamically.

I know that the folder layout isn't perfect, but I didn't want to have to change all the imports everywhere to get this going.  I also delete the transpiled output often to ensure that only files which exist in src/ will be in bin/ and lib/.  The case that I'm concerned is that I have src/lib/filea.js which gets transpiled to lib/filea.js.  If I then delete src/lib/filea.js, babel will not know to delete lib/filea.js.


package.json
```diff
@@ -3,14 +3,23 @@
...
+    "pretest": "npm run compile",
     "test": "test/runtests.sh"
   },
+  "babel": {
+    "sourceMaps": true,
+    "optional": [
+      "runtime"
+    ],
+    "stage": 1
+  },
```
jonasfj: Why put it here? If we were publishing this to npm we wouldn't require consumers to use babel... Isn't that what this is for?
I suspect it's supported here for people who load babel modules using browserify w. transformer, or in a  node project...

I might be wrong here... You could certainly keep using `.babelrc`, but I'm not sure what is right...

jonasfj:
`r+` for sure... this good we should this everywhere.. :)

-----
I was hoping we could have a simpler folder layout:
```
lib/ (compiled files)
src/ (source files)
```
So no `src/lib/` or `src/bin/`... This is not needed now... it's a small thing to move later..
I was thinking that when we have component loader, we'll just have one bin file `src/server.js`, which will be compiled to `lib/server.js`. And then we'll launch it as `node lib/server.js <profile> <component>`, where `<component>` is a component defined in the loader...

Then we only have one bin file `server.js` we can put source map support there too:
```js
require('source-map-support').install({
  handleUncaughtExceptions: true
});

var load = base.loader({....});

if (!module.parent) {
  load({profile: process.argv[2]}, process.argv[3]); // just illustrative haven't gotten to that PR yet :)
]

module.exports = load;
```

Also you probably need `--require source-map-support/registe` in `mocha.opts` for source maps to work... You should test that they work...

package.json
jhford: I like having a single place where all the configuration for a project goes.  it's also nice that it's not a hidden file.

jonasfj: Okay, sounds good... As discussed in the other PR maybe in the future this is something we can load from an npm module...

jonasfj:
@jhford, did this land some how or???
If landed we should close this..

Note: I'm slowly working on getting queue and auth transpiling... slowly because it works...

PR #49, opened 2015-09-30
Bias to influx

jhford:
This will be easier to review after #48 lands...  Basically, we should be submitting the bias that we used to calculate the price along with the rest of the bids to influx.


jonasfj:
`r+`, note we should probably learn to file PRs as merge this into another branch...
I just did compare as I would for a PR merging `bias-to-influx` into `static-transpile`

jhford:
landed as bf642e3

PR #52, opened 2015-11-30
Fix SSH Key rotation so that we change the KeyPair with a new awsInstancePubkey setting

jhford:
Basically add a sha256[0:7] of the public key to the KeyPair and parse that out when enumerating keys.
This also factors out the KeyPair parsing/generating stuff into nice functions.  We can't delete
old keys until we know that there are no instances using the old KeyPair.  I guess we could go through
all instances and make sure that all keys which start with the prefix but don't have a running instance
or the ssh pubkey hash of the current pubkey are deleted... not too important though

### DIFF
--- a/lib/aws-manager.js
+++ b/lib/aws-manager.js
@@ -4,6 +4,7 @@ let debug = require('debug')('aws-provisioner:aws-manager');
 let shuffle = require('knuth-shuffle');
 let taskcluster = require('taskcluster-client');
 let series = require('./influx-series');
+let keyPairs = require('./key-pairs');
 
 const MAX_ITERATIONS_FOR_STATE_RESOLUTION = 20;
 
@@ -671,7 +672,7 @@ class AwsManager {
     for (let region of this.ec2.regions) {
       for (let reservation of instanceState[region].Reservations) {
         for (let instance of reservation.Instances) {
-          let workerType = instance.KeyName.substr(this.keyPrefix.length);
+          let workerType = this.parseKeyPairName(instance.KeyName).workerType;
           // XXX let filtered = objFilter(instance, this.filters.instance);
           let filtered = instance;
           filtered.Region = region;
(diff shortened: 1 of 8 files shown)


lib/aws-manager.js
```diff
@@ -1081,6 +1082,27 @@ class AwsManager {
   }
 
   /**
+   * wrapper for brevity
+   */
+  createPubKeyHash () {
```
jonasfj: Method is never called

lib/aws-manager.js
```diff
@@ -1081,6 +1082,27 @@ class AwsManager {
...
+   */
+  createKeyPairName (workerName) {
+    return keyPairs.createKeyPairName(this.keyPrefix, this.pubKey, workerName);
+  }
+
+  /**
+   * wrapper for brevity
+   */
+  parseKeyPairName (name) {
+    return keyPairs.parseKeyPairName(name);
```
jonasfj: Maybe we can skip the wrapper here... :)

lib/key-pairs.js
```diff
@@ -0,0 +1,51 @@
...
+  return prefix + workerName + ':' + module.exports.createPubKeyHash(pubKey);
+};
+
+/**
+ * Parse a KeyPair name into an object with a prefix, workerType and
+ * keyHash properties on the returned object
+ */
+module.exports.parseKeyPairName = function (name) {
+  assert(name);
+  let parts = name.split(':');
```
jonasfj: You need to:
 1) test that `prefix` doesn't contain `:`
 2) add a `:` between in `prefix + ":" + workerName + ':' + module.exports.createPubKeyHash(pubKey);`

Otherwise this won't work... I suspect this currently relies on the prefix containing a colon.

jhford: it does.  I'll add this check and do it this way, but I'm going to have it automatically trim a single trailing : if it's on the prefix

jonasfj:
a few nits... but generally this looks good... The one where `createKeyPairName` and `parseKeyPairName` relies on `prefix` ending with a `:` is probably worth fixing.

The others are just suggestions, feel free to ignore :)

jhford:
Good idea, I've changed how I'm doing the keyPrefix to what you've suggested!

jonasfj:
Awesome, remember to push it too... :)

Anyways, feel free to roll it out with that fix, seems like this will generally do good.

PR #53, opened 2016-01-26
pass throw deadman's snitch api

jhford:
because status.taskcluster.net really should have this info


lib/routes/v1.js
```diff
@@ -710,6 +712,34 @@ api.declare({
 
 api.declare({
   method: 'get',
+  route: '/backend-status',
+  name: 'backendStatus',
+  title: 'Backend Status',
+  description: [
```
jonasfj: Adding `stability: 'experimental',` would have been nice... (but not critical)

jonasfj:
@jhford, pretty sweet.. `r+`

lib/routes/v1.js
jonasfj: Actually, I guess that is the default stability level... so never mind :)

PR #55, opened 2016-02-22
Remove multi-region-aws-promised -WIP

jhford:
This is only a work in progress so far, but I wanted @jonasfj to take a look at it before I go too much further.  I tried to keep the data structures the same, but I did remove the multi-region library.  I did this because my diagnostic to see if the multi-region library was busted was changing behaviour, so I decided to just rip it out.

With the diagnostic patch, I was showing processes crashing but not the way I thought they would.

cc @selenamarie fyi


jhford:
DO NOT MERGE THIS YET!  It's not ready

bin/provisioner.js
```diff
@@ -1,7 +1,7 @@
 let debug = require('debug')('aws-provisioner:bin:provisioner');
 let base = require('taskcluster-base');
 let provision = require('../lib/provision');
-let Aws = require('multi-region-promised-aws');
+let awsSdk = require('aws-sdk-promise');
```
jonasfj: nit, but everywhere else we do: `let aws = require('aws-sdk-promise');`
Just `aws` as variable...

lib/aws-manager.js
```diff
@@ -81,142 +82,164 @@ class AwsManager {
...
+    this.__availableAZ = availableAZ;
+    this.__pricing = allPricingHistory;
+    this.__apiState = apiState;
+    this.__deadState = deadState;
 
-    // We store the last iteration's state
-    this.__previousApiState = this.__apiState;
-    this.__apiState = this._classify(livingInstances, filteredSpotRequests.good);
-    this.__deadState = this._classify(deadInstances, deadSpotRequests);
+    await Promise.all(stalledKills);
```
jonasfj: Because async() evaluates async, `stalledKills == []` always.. the place where you push entries to it is never reached before the `_.forEach` is done.. as `_.forEach` is sync...

lib/aws-manager.js
```diff
@@ -81,142 +82,164 @@ class AwsManager {
...
-      });
-      debug('fetched living spot request state');
+    let stalledKills = [];
+    let availableAZ = {};
+    let allPricingHistory = {};
+
+    // It would be better to do it by classification than by region for
+    // better concurrency, but this is easier and not too slow considering
+    // the 75s iteration frequency
+    _.forEach(this.ec2, async (obj, region) => {
```
jonasfj: Here we have a high risk of dropping the ball, well dropping the promise :)

You'll want to do:
```js
await Promise.all(_.map(this.ec2, async (ec2, region) => {
   ....
}));
```

This way anything that goes wrong in here will propagate up... `_.forEach(, ...async ()` is always a footgun, a major one...
Technically, this could work fine, if you are confident that there are no bugs in the async function, but we don't have to be confident about that... So let's not even try...

nit: also using `ec2` instead of `obj` is much nicer :)

lib/aws-manager.js
```diff
@@ -81,142 +82,164 @@ class AwsManager {
...
+      let response = await Promise.all([
```
jonasfj: To makes this more readable I strongly propose we unpack... We can do this in two ways:
```js
let [
  livingInstances,
  livingSpotRequests,
  deadInstances,
  deadSpotRequests,
  availabilityZones,
  spotPriceHistory,
] = await Promise.all([
  ....
])
```

You can indeed make your life a tiny bit simpler with:

```js
let [
  livingInstanceReservations,
  livingSpotRequests,
  deadInstances,
  deadSpotRequests,
  availabilityZones,
  spotPriceHistory,
] = await Promise.all([
        // Living instances
        obj.describeInstances({
          Filters: [
            {
              Name: 'key-name',
              Values: [this.keyPrefix + '*'],
            },
            {
              Name: 'instance-state-name',
              Values: ['running', 'pending'],
            },
          ],
        }).promise().then(_.property('data.Reservations')),
]);
```

Or if you want to make your life very simple you can do:
```js
let [
  livingInstances,
  livingSpotRequests,
  deadInstances,
  deadSpotRequests,
  availabilityZones,
  spotPriceHistory,
] = await Promise.all([
        // Living instances
        obj.describeInstances({
          Filters: [
            {
              Name: 'key-name',
              Values: [this.keyPrefix + '*'],
            },
            {
              Name: 'instance-state-name',
              Values: ['running', 'pending'],
            },
          ],
        }).promise()
        .then(_.property('data.Reservations'))
        .then(r => _.map(r, _.property('Instances')))
        .then(_.flatten),
])
```

I think the last one might be over kill, but using `.then(_.property('data.Reservations'))` is reasonably sane....

lib/aws-manager.js
```diff
@@ -81,142 +82,164 @@ class AwsManager {
...
+      for (let reservation of response[0].data.Reservations) {
+        for (let instance of reservation.Instances) {
+          let workerType = this.parseKeyPairName(instance.KeyName).workerType;
+          // Maybe use objFilter here
+          let filtered = instance;
+          filtered.Region = region;
+          filtered.WorkerType = workerType;
+          apiState.instances.push(filtered);
+        }
+      };
```
jonasfj: Rather than `apiState.instances.push(filtered)` let's see if we can't do `.map` instead..

```js
apiState.instances = _.flatten(_.map(response[0].data.Reservations, _.property('Instances'))).map(instance => {
  instance.WorkerType = this.parseKeyPairName(instance.KeyName).workerType;
  instance.Region = region;
  return instance;
});
```
If we use the `.then(_.property('data.Reservations'))` proposed earlier this becomes: 
```js
apiState.instances = _.flatten(_.map(livingResevations, _.property('Instances'))).map(inst => {
  inst.WorkerType = this.parseKeyPairName(instance.KeyName).workerType;
  inst.Region = region;
  return inst;
});
```

I propose this because it's now very clear that `apiState.instances` is set here... and you don't get the idea that it's partially populated before this...

jonasfj: **Sorry, I'm wrong...**

You can't do this because the call for each region, adds to this list in parallel..

lib/aws-manager.js
jonasfj: I propose splitting this into 4 methods...
```js
  async _fetchLivingInstances(region) {
    // Get resevations
    let resevations = await this.ec2[region].describeInstances({
      Filters: [
        {
          Name: 'key-name',
          Values: [this.keyPrefix + '*'],
        },
        {
          Name: 'instance-state-name',
          Values: ['running', 'pending'],
        },
      ],
    }).promise().then(_.property('data.Reservations'))
    
    // For each instance add WorkerType and Region as properties
    return _.flatten(_.map(resevations, _.property('Instances'))).map(i => {
      i.WorkerType = this.parseKeyPairName(instance.KeyName).workerType;
      i.Region = region;
      return i;
    });
  }
```

Then instead of this long `_.forEach` I propose we make some simple `_.map` followed by `_.flatten`
```js
[
  apiState.instances,
  apiState.requests,
  ...
] = await Promise.all([
  Promise.all(_.map(this.ec2, region => this._fetchLivingInstances)).then(_.flatten),
  Promise.all(_.map(this.ec2, region => this._fetchLivingRequests)).then(_.flatten),
  ...
]);
```
Everything still runs in concurrently, but we've split it into smaller methods...
Each method returns a list, we avoid having to do the `apiState.instances.push(...)` by accepting that each method returns a list which we then `_.flatten` afterwards...

We then use unpacking from `await Promise.all` go  the result assigned to the correct `apiState` properties...

I realize that `availableAZ` and `allPricingHistory` won't fit this pattern nicely...
So we can either just hack them at the end... and disregard the return value when unpacking...
Or, probably better, we use `_.spread(_.assign)` instead of `_.flatten`...

This is all more functional, as in we returns values, and merge values into new values... as oppose to being imperative where we transform objects and arrays by adding things to them.

Generally, this functional-style programming will save you a lot pain and keep things simple.

lib/aws-manager.js
```diff
@@ -231,70 +254,24 @@ class AwsManager {
...
+      let price = parseFloat(pricePoint.SpotPrice, 10);
+      let zone = pricePoint.AvailabilityZone;
+
+      // Remember that we only want to consider available zones
+      if (zones.includes(zone)) {
+        if (!pricing[type]) {
+          pricing[type] = {};
+        }
+        if (!pricing[type][zone] || pricing[type][zone] < price) {
+          pricing[type][zone] = price;
```
jonasfj: I think this is a comment for another time, but you would be a lot better off with:
```js
pricing[type] = [
 {zone: ..., price: ...},
]
```
Well, it depends on how it is used later... if you're looking for the cheapest zone (which I suspect, then having an array makes it easier to some form of max function... I'm sure lodash has one to select max object from a list...

lib/aws-manager.js
```diff
@@ -309,67 +286,50 @@ class AwsManager {
    * This function returns an object with the spot requests sorted
    * into these buckets.
    */
-  _filterSpotRequests (spotReqs) {
-    let data = {
-      good: {},
-      stalled: {},
-    };
-
+  _spotRequestStalled (sr) {
```
jonasfj: ++ on complexity reduction !!! :)

lib/aws-manager.js
```diff
@@ -1122,39 +1025,30 @@ class AwsManager {
...
-          }
-          throw err;
-        }
+      // Since we're using a filter to look for *only* this
+      // key pair, the only possibility is 0 or 1 results
+      if (!keyPairs.KeyPairs[0]) {
+        await this.ec2[region].importKeyPair({
+          KeyName: keyName,
+          PublicKeyMaterial: this.pubKey,
+        });
```
jonasfj: Missing `.promise()` that's not so good :)

lib/aws-manager.js
```diff
@@ -1122,39 +1025,30 @@ class AwsManager {
...
 
-    let res = await this.ec2.describeKeyPairs.inRegions(this.ec2.regions, {
-      Filters: [
-        {
-          Name: 'key-name',
-          Values: [keyName],
-        },
-      ],
-    });
+    for (let region of _.keys(this.ec2)) {
```
jonasfj: again, ++ for making this simpler... I mean really this is actually simpler because the response format is less convoluted...

Nit:
```js
await Promise.all(_.map(this.ec2, async (ec2, region) => {
   let keyPairs = await ec2.describeKeyPairs({
     ...
   }).promise();
   
   if (keyPairs.KeyPairs[0]) {
     return;
   }

   await ec2.importKeyPair({
     ....
   }).promise();
}));
```
This would give you max concurrency... and a slight complexity reduction as you won't have to do `this.ec2[region]` anymore... because `_.map` gives us both value and key.
(yes, lodash ordering of `(value, key) => {` in map, forEach etc... is driving me insane :)

lib/aws-manager.js
```diff
@@ -1247,7 +1125,9 @@ class AwsManager {
 
     let perRegionKills = {};
 
-    for (let region of this.ec2.regions) {
+    let killPromises = [];
+
+    for (let region of _.keys(this.ec2)) {
```
jonasfj: We should convert this into `Promise.all(_.map(this.ec2` and avoid the `perRegionKills` data structure...

I've proposed something like this before, but here it is again. The `SmartFilter` a better name is welcome:
```js
/**
 * A [].filter-like method that takes a list and an object of conditions.
 *
 * SmartFilter([{...}, {...}, ...list of objects], {
 *   'a.property.path': value // a value that must be matched
 *   'property': val => ... // function that evaluates value at path...
 *   'sub.prop': [1, 2, 3]  // a list of allowed values, must be one of  
 * });
 *
 * Each property in the condition object is a condition. All conditions must
 * be satisfied. Each condition is a key, value pair of the condition object.
 * The key is an object path, the value is a value that must be matched, a
 * function that must return true, or a list which must contain the value.
 */
SmartFilter = (items, filter) => {
  return _.filter(items, item => {
    return _.every(filter, (val, key) => {
      if (val instanceof Array) {
        return _.includes(val, _.get(item, key));
      } else if (val instanceof Function) {
        return val(_.get(item, key));
      } else {
        return _.isEqual(_.get(item, key), val);
      }
    });
  });
};
```

```js
await Promise.all(_.map(this.ec2, async (ec2, region) => {
  let instanceIds = SmartFilter(this.__apiState.instances, {
    'WorkerType: name,
    'State.Name': states, // this is a one-of, see the implementation of SmartFilter
    'Region': region,
  }).map(_.property('InstanceId'));

  let requests = [];  
  .... // smilar hacks... probably using _.flatten to merge the two lists.. As you'll call SmartFilter twice

  await this.killCancel(region, instanceIds, requests);
}));
```

I know `SmartFilter(items, conditions)` might take a few seconds to understand.. And yes, the code isn't much shorter... In fact it's far more in-efficient because we iterate over all instances once for each region...

But all of that doesn't matter.. The code is much more declarative and functional. We declare how we want the list of instances filtered, rather than doing the actual filtering.

@jhford, please seriously consider this `SmartFilter` approach, it'll make so many things much clearer... I'll be happy to discuss it with you... I think it can be used many places in the provisioner code based, because we're operating on lists of objects, and we want to filter them..

(but manual filtering is boring, error prone, and **hard to read**, so let's try to be declarative).

jonasfj:
@jhford These comments are not because this is bad at all.

Github says: `Showing  6 changed files  with 277 additions and 445 deletions.`

So you've removed a complicated dependency and still reduced the number of lines of code!
That's impressive.

bin/provisioner.js
jhford: heh, i just didn't remember if I had another variable.  changed

lib/aws-manager.js
jhford: I like the idea of SmartFilter, but I think we can get just as clear results with what I've changed to doing:

```javascript
      let instances = this.__apiState.instances.filter(instance => {
        return instance.WorkerType === name && states.includes(instance.State.Name); 
      });
```

jhford:
@jonasfj I've addressed a bunch of your comments!  Let me know what you think

lib/aws-manager.js
jhford: yah, lets worry about that later.

lib/aws-manager.js
jhford: :)

lib/aws-manager.js
jhford: moved this into the per-region code.  The API call shouldn't really slow anything down in any meaningful way by doing it there and avoids this situation altogether.

lib/aws-manager.js
jhford: did this

lib/aws-manager.js
jhford: I think this is probably something we should do, but I think it's a little too much for this pull request.

lib/aws-manager.js
jhford: ugh.  i hate that this library requires you to call .promise().  Oh well.

lib/aws-manager.js
jhford: fixed this.

jhford:
Oh, and this is roughly working on my local machine.

lib/aws-manager.js
```diff
@@ -81,142 +82,162 @@ class AwsManager {
...
+        ec2.describeSpotPriceHistory({
+          StartTime: pricingStartDate,
+          Filters: [
+            {
+              Name: 'product-description',
+              Values: ['Linux/UNIX'],
+            },
+          ],
+        }).promise(),
+      ]);
```
jonasfj: I still propose refactoring this into methods and combine the result with `_.flatten` and `_.assign` like:
```js
[
  apiState.instances,
  apiState.requests,
  ...
] = await Promise.all([
  Promise.all(_.map(this.ec2, region => this._fetchLivingInstances)).then(_.flatten),
  Promise.all(_.map(this.ec2, region => this._fetchLivingRequests)).then(_.flatten),
  ...
]);
```
See comment: https://github.com/taskcluster/aws-provisioner/pull/55#discussion_r53674042

----
It's probably not blocking, but seriously consider it.

lib/aws-manager.js
jonasfj: agree!

lib/aws-manager.js
```diff
@@ -1168,35 +1060,27 @@ class AwsManager {
...
 
-    for (let region of regions) {
-      try {
-        await this.ec2.deleteKeyPair.inRegion(region, {
+      // Since we're using a filter to look for *only* this
+      // key pair, the only possibility is 0 or 1 results
+      if (keyPairs.data.KeyPairs[0]) {
+        await ec2.deleteKeyPair({
           KeyName: keyName,
         });
```
jonasfj: Missing `.promise()`

lib/aws-manager.js
```diff
@@ -1168,35 +1060,27 @@ class AwsManager {
...
+        ],
+      }).promise();
 
-    for (let region of regions) {
-      try {
-        await this.ec2.deleteKeyPair.inRegion(region, {
+      // Since we're using a filter to look for *only* this
+      // key pair, the only possibility is 0 or 1 results
+      if (keyPairs.data.KeyPairs[0]) {
+        await ec2.deleteKeyPair({
```
jonasfj: you are missing `.promise()`

lib/aws-manager.js
```diff
@@ -1245,50 +1120,26 @@ class AwsManager {
...
+        return instance.WorkerType === name && _.includes(states, instance.State.Name); 
+      });
 
-    for (let instance of this.__apiState.instances) {
-      if (instance.WorkerType === name && states.includes(instance.State.Name)) {
-        perRegionKills[instance.Region].instances.push(instance.InstanceId);
-      }
-    }
+      // _.concat just doesn't seem to work, so I'm doing this a slightly
+      // different way.  _.concat is showing as undefined :/
```
jonasfj: I'm no `.concat` fan either.. regardless if it is `[].concat` or `_.concat`..

I suggest:
```js
let requests = _.flatten([
  this.__apiState.requests.filter(requests => {
    ....
  }),
  this.__internalState.filter(requests => {
    ....
  }),
]);
```

`_.flatten([1, [2], [3,4], [[5]]]) == [1, 2, 3, 4, [5]]` is pretty nice in that it creates a new list... rather than concatenating... And it only removes one level of arrays...

lib/aws-manager.js
```diff
@@ -1300,37 +1151,18 @@ class AwsManager {
...
-      } catch (err) {
-        debug('error terminating instances in %s', region);
-        debug(err);
-        if (err.stack) {
-          debug(err.stack);
-        }
-        throw err;
-      }
-    }
+    let promises = [];
```
jonasfj: You don't need this...

Also don't do default values like this... just do `async killCancel (region, instances = [], requests = []) {`

lib/aws-manager.js
```diff
@@ -1300,37 +1151,18 @@ class AwsManager {
...
-      } catch (err) {
-        debug('error cancelling spot requests in %s', region);
-        debug(err);
-        if (err.stack) {
-          debug(err.stack);
-        }
-        throw err;
-      }
-    }
+    //if (i.length > 0) {
```
jonasfj: You do need this, right?

jonasfj:
I'm okay with you doing: https://github.com/taskcluster/aws-provisioner/pull/55#discussion_r53674042
In a different PR, but seriously consider it... Splitting into methods would simplify this a lot...

lib/aws-manager.js
jonasfj: >
> I like the idea of SmartFilter, but I think we can get just as clear
> results with what I've changed to doing:

Yeah, it needs a better name than "SmartFilter"... maybe just "filter".
The idea with it was that we could use it for a bunch of things...

There are many places where you need to filter in the manner you do here...
Specially when looking at either results from ec2, or apiState...

I think I've said it before... but I would have merged all entries in
apiState properties into one array...
Even the different types... and then I would have .type = 'spot-request' or
.type = 'instance', etc...

That's a larger thing.. but I  suspect you could get a nice design.
Basically treating apiState = [...] as a nosql database, that you just
query..

Anyways, feel free to ignore it... but you have data that you look at in
different ways, a big hammer is probably useful :)

--
Regards Jonas Finnemann Jensen.

2016-02-23 3:29 GMT-08:00 John Ford <notifications@github.com>:

lib/aws-manager.js
jhford: heh, yes, there was also a bug where I left off the .promise() method... I hate that we need to call the .promise() method... I'd so much rather have x + 'Async' pattern.

jhford:
cool.  This is mostly working for me locally.  I think the only thing that's going to prove it completely is some time in production.  @jonasfj do you support me deploying this to production as it exists now?  My plan is to deploy morning tomorrow my time then monitor it for the day, reverting if things break.

jonasfj:
yeah, roll it, but please consider following up with https://github.com/taskcluster/aws-provisioner/pull/55#discussion_r53674042

The current code is still not easily readable and my confidence in it isn't too high... Higher than before :)
It's much easier to review and trust the code if it's better structured.. Small functions, more functional approach is a good way to do that.

Lol, soon I'll probably be preaching erlang and haskell... What's happening to me!

PR #56, opened 2016-02-23
remove the scope required for the listWorkerTypes method

jhford:


jonasfj:
`r+`... I suspect you already landed it...

PR #57, opened 2016-02-24
Refuse to allow insertion of bad workertype definitions

jhford:
@jonasfj how do you feel about this?  I should still probably put a try/catch around the requestSpotInstances and check if the failure to request a spot instance is because of a bad ami id, and if so, ignore the error and maybe send a pulse message or something?

fyi @petemoore and grenade, but I don't know his github.

### DIFF
--- a/lib/aws-manager.js
+++ b/lib/aws-manager.js
@@ -1037,12 +1037,20 @@ class AwsManager {
     assert(this.__availableAZ[bid.region].includes(bid.zone),
         'will not submit spot request in an unavailable az');
 
-    let spotRequest = await this.ec2.requestSpotInstances.inRegion(bid.region, {
-      InstanceCount: 1,
-      Type: 'one-time',
-      LaunchSpecification: launchInfo.launchSpec,
-      SpotPrice: bid.price.toString(),
-    });
+    try {
+      let spotRequest = await this.ec2.requestSpotInstances.inRegion(bid.region, {
+        InstanceCount: 1,
+        Type: 'one-time',
+        LaunchSpecification: launchInfo.launchSpec,
+        SpotPrice: bid.price.toString(),
+      });
+    } catch (err) {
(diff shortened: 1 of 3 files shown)


lib/aws-manager.js
```diff
@@ -1037,12 +1037,20 @@ class AwsManager {
...
+    try {
+      let spotRequest = await this.ec2.requestSpotInstances.inRegion(bid.region, {
+        InstanceCount: 1,
+        Type: 'one-time',
+        LaunchSpecification: launchInfo.launchSpec,
+        SpotPrice: bid.price.toString(),
+      });
+    } catch (err) {
+      if (err.code === 'InvalidAMIID.NotFound') {
+        debug(err.message + 'in ' + bid.region);
```
jonasfj: probably using `console.log` I'm starting to think we should just use that for things that aren't debug messages...

jonasfj: nit...

lib/routes/v1.js
```diff
@@ -59,6 +60,97 @@ let api = new base.API({
...
+    }));
+  }));
+
+  // Finally, let's verify that the image for this workerType exists in EC2
+
+  if (reasons.length > 0) {
+    debug('Found errors with ' + workerTypeName); 
+    let e = new Error('Refusing to create an invalid worker type');
+    e.reasons = reasons;
+    throw e
```
jonasfj: @jhford, this okay... But really you shouldn't communicate it this way...

`validateWorkerType` should just return a string message if there is an error, or a `null` if there is no error..
Or something like this, you generally shouldn't use exceptions to communicate return value.

jonasfj: Particularly because an unhandled exception is more of a 500 and than a 400

petemoore:
Thanks for taking care of this @jhford :+1:

PR #59, opened 2016-03-01
two minute timeout on api state calls

jhford:
@jonasfj, i've tested this locally with a 1ms delay instead of a 120s delay and it does properly handle the timeout


jonasfj:
`r+`... Please consider giving ec2 a logger as well...

jhford:
yep! that's the plan

PR #60, opened 2016-03-01
Updates

jhford:
Here's some stuff that I would like to test and deploy in the morning.


package.json
```diff
@@ -13,34 +15,36 @@
     "aws-sdk": "^2.2.37",
     "aws-sdk-promise": "0.0.2",
     "babel": "^5.8.20",
-    "bluebird": "^2.9.34",
+    "babel-runtime": "^6.6.1",
     "commander": "^2.6.0",
     "debug": "^2.2.0",
-    "irc": "^0.4.0",
-    "kexec": "^1.2.0",
+    "kexec": "^2.0.2",
```
jonasfj: I think you can drop it...

jonasfj:
Wow, that's a lot of cleanup :)

`r+`...

I feel so guilty for not doing this in things I own....

jonasfj:
I checked every change a second time... and from the github syntax it's easy to see that you got them all right...

PR #63, opened 2016-03-31
Bug 1261024 - Align Create/Get WorkerType schemas in aws provisioner

petemoore:
This fixes `get-worker-type-response.yml` title property, and makes `region` a required property in `create-worker-type-request.yml`.


petemoore:
Thanks!

PR #65, opened 2016-04-07
Bug 1261000: update README, get tests running for me

djmitche:
The package.json change to eslint is to prevent 2.3.0 being installed,
which is incompatible with some other package.

I was surprised to find that the AWS key information is not actually used.  I suspect that means a lot of this code isn't being tested by the unit tests.  Anyway, this is a start.


README.md
```diff
@@ -1,5 +1,52 @@
...
+To run the provisioner locally, you will need a similar set of configuration.
+Then run
+
+```
+npm compile && DEBUG=* node lib/server.js development
+```
+
+Note that this is not usually the best way to test the provisioner.  Be careful
+that you manually destroy any resources the provisioner creates while it is
+running (EC2 instances or spot bids).
```
petemoore: Thanks for working on this @djmitche! I like what I see! Could you also add:

* a section on how to deploy the provisioner (e.g. if it happens automatically with a push to master, or if you need to push to heroku) and what environment variables need to be configured when setting this up
* information about what objects the provisioner creates in AWS, so people know what to clean up, and e.g. if someone were to set up a CI running in travis, they'd know how to do it in a way that they wouldn't pollute our aws environment

I think those are the only two pieces I miss. Thanks again for picking this up!

TODO.md
```diff
@@ -1,28 +0,0 @@
...
-* Provide facilities for copying AMIs between regions
-* Api endpoints for
-  * cancelling spot requests
-  * checking on status of spot request
-  * metrics on current state
-* insert into influx when the hello endpoint is hit
-* Ensure that UserData is valid JSON
-* Write tests for all the provisioning logic
-* Wait up to a certain amount of time before assuming that
-  a spot request is cancelled
```
petemoore: Out of interest, why was this deleted? Is this tracked somewhere else now?

README.md
djmitche: I don't know the answers to either of those questions :)

TODO.md
djmitche: Some of these are no longer the plan, some are done, and some I don't recognize.  In general, I think we already have a long list of important TODO's in the Bugzilla component, so we can rely on that.

README.md
djmitche: I suppose I could look it up in Heroku :)

djmitche: @petemoore updated

petemoore:
nice, thanks!

PR #66, opened 2016-04-07
Bug 1261911: remove unused/testing endpoints

djmitche:


src/api-v1.js
```diff
@@ -646,48 +646,6 @@ api.declare({
 
 api.declare({
   method: 'get',
-  route: '/worker-type/:workerType/launch-specifications',
```
jhford: Please leave this.  It's marked as experimental and I do use it.

src/api-v1.js
```diff
@@ -779,51 +735,3 @@ api.declare({
     uptime: process.uptime(),
   });
 });
-
-api.declare({
-  method: 'get',
-  route: '/backend-status',
```
jhford: Please do not delete this.  It is supposed to be used to get information into status.taskcluster.net.  If that's not where the data for the aws-provisioner graph is coming from, then the graph is measuring the wrong thing.

djmitche: It's not -- status comes from the ping endpoint.  I'll file a bug to document this endpoint and its schemas, and to update the status tool (not status.tc.n) to use that endpoint

djmitche: https://bugzilla.mozilla.org/show_bug.cgi?id=1261912

djmitche:
ok, updated

PR #67, opened 2016-04-07
Bug 1198391: don't require scopes for workerType state

djmitche:
This changes the state(<workerType>) endpoint to just return summary
information about the workerType, not restricted by scopes.  With this
change, the UI can helpfully display current capacity information to
anyone, regardless of scopes.

@pmoore too..


djmitche:
(the tools UI currently doesn't call this method; once this lands I'll rework that code to use this method, then delete the awsState() method)

djmitche:
(the latter in bug 1261910)

src/api-v1.js
```diff
@@ -740,26 +740,64 @@ api.declare({
...
+    maxCapacity: workerType.maxCapacity,
+  };
+
+  let capacities = {};
+  workerType.instanceTypes.forEach(instanceType => {
+    capacities[instanceType.instanceType] = instanceType.capacity;
+  });
+
+  workerState.instances.forEach(instance => {
+    // count every AWS state other than 'running' as 'pending'
```
jhford: we should only ever have running and pending.  Let's ignore instances that aren't in either of these states.

src/api-v1.js
```diff
@@ -740,26 +740,64 @@ api.declare({
   route: '/state/:workerType',
   name: 'state',
   title: 'Get AWS State for a worker type',
-  scopes: [['aws-provisioner:view-worker-type:<workerType>']],
+  output: 'state-response.json#',
```
jhford: Instead of getting rid of the scope-protected endpoint, I'd much rather that we add this as a new endpoint called `/simple-state` or something like that.  It would be a really unfortunate outcome to lose the ability to see and link to ec2 console for instances and other things in the UI that the current implementation of `/aws-state/` and `/state/` allow us to have.

As this method is implemented right now, I think it's great!

src/worker-type.js
```diff
@@ -77,12 +77,15 @@ let WorkerType = base.Entity.configure({
     canUseSpot: base.Entity.types.JSON,
     /* This dictionary describes which instance types that this workerType
      * can run on as well as type-specific information.
-     * This is a dictionary in the shape:
+     * This is a list of objects:
      * {
-     *   'c1.small': {
+     *     instanceType: 'm3.7xlarge',
```
jhford: thanks for updating the docs.

jhford:
Looks good!  Let's implement the scope-less method as a second method and leave the `/state/` method there.  `/state/` might not be generally useful, but I do like having it available.  It would also be neat if someone who has credentials for `/state/` loaded that instead of this method.  I think that we could probably open `/state/` up to anyone who can modify the worker type definition safely.

src/api-v1.js
djmitche: How about adding a new `/state/:workerType/ec2-objects` / `ec2Objects(..)` endpoint for that purpose?  From a naming perspective, "state" is typically some basic summary information, so I feel like `simple-state` is really just a workaround for `state` already being taken.

jhford: what about `/abridged-state/:workerType`, `/state-summary/:workerType` or `/capacity-info/:workerType`?  it's not ec2 objects being returned by that method, rather an internal format that's quite different to the ec2 objects.  In my mind, the 'real' call is the authenticated one and this one should only be used by those lacking the credentials to load the auth'd call.

djmitche: I think the calls are two different things -- if you want to know roughly "how this worker type is doing" (meaning, is it pegged at its max capacity? is it busily spinning up new instances?), you call the status endpoint.  If you want to know exactly which resources are associated with it, you call the ec2-objects (or maybe resources?) endpoint.  I expect that the former is going to be much more useful to everyone -- it might appear in treeherder, for example, as an indicator of current load; or be sampled periodically to create some kind of "burn rate" expense dashboard.

I strongly dislike the "internal format" idea -- an API shouldn't expose internal details, and whatever that returns is, by definition, an external representation.  And that representation should make sense to someone not familiar with the implementation.  In this case, it's pretty sensible: there are a set of instances in either the pending or running state, and a set of requests which have not yet been associated with an instance.  I'd actually prefer to put that in a single list of EC2 resources, some of which just have 'null' for the instanceId, indicating that they are un-fulfilled requests.

djmitche:
OK, I moved the endpoints around a little, and renamed them.  Now there is

* `GET /list-worker-types` (listWorkerTypes)
* `GET /worker-type/:workerType` (workerType)
* `PUT /worker-type/:workerType` (createWorkerType)
* `POST /worker-type/:workerType/update` (updateWorkerType)
* `DELETE /worker-type/:workerType` (removeWorkerType)
* `GET /worker-type/:workerType/capacity` (capacity)
* `GET /worker-type/:workerType/resources` (resources)
* `GET /worker-type/:workerType/launch-specifications` (getLaunchSpecs)

.. and some secret stuff and miscellany.  I think that the URL forms make more sense here, and I took your advice to name it "capacity".

jhford:
There's no need to rename any existing endpoints.  This is about adding a new endpoint, let's go with `/state-summary/:workerType` as the new endpoint name and not rename anything.

djmitche:
I'm only renaming the endpoint I haven't commited yet.  I think it makes a lot more sense for all of the endpoints that describe a single worker-type to start with `/worker-type/:workerType/`

jhford:
according to your comment, you're also renaming the /state/ method.

djmitche:
Well, "rename" or "remove in favor of".  Look at it as removing the `state()` method and adding `capacity()` (without scopes) and `resources()` (requiring scopes),  Remember that the existing `state()` method isn't used yet -- tools still uses `awsState()`.  So this isn't a breaking change.

jhford:
I don't think capacity and resources are very specific names.  resources will have capacity information and capacity is a measure of resources.  At that point, we're trading intentionally vague naming for ambiguous information.

Looking at the code again, the downside to the approach here, as opposed to having the scope-less call be a whitelisted subset of info from the scope-ful set is that there are two very different codepaths to read the information from the api call.  It also means that we can only see capacity totals and not per-region or availability zone capacity, which can be really useful to unauthenticated users as well.

schemas/capacity-response.yml
```diff
@@ -0,0 +1,46 @@
...
+      required: [instances, capacity]
+    maxCapacity:
+      title: Maximum capacity of the workerType
+      type: integer
+  additionalProperties: false
+  required:
+    - running
+    - pending
+    - spotReq
+    - maxCapacity
```
jonasfj: Is there any reason for structuring this like this rather than having:
`runningCapacity, runningInstances, pendingCapacity, pendingInstances, requestedSpotInstances, requestedSpotCapacity`

Nesting objects for structure is rarely desirable... I've long regretted `task.metadata`, the members should just have been top-level properties. (just figured I would warn you).

schemas/resources-response.yml
```diff
@@ -0,0 +1,52 @@
...
+          region:
+            type: string
+            description: AWS Region
+          zone:
+            type: string
+            description: AWS Availability Zone
+          state:
+            type: string
+            description: |
+              Resource state; "running" indicates an instance able to execute tasks.
```
jonasfj: I bet this an `enum` it would be really nice for the consumer of this API if the possible values was documented and validated...

Even if the `description` comment says that we retain the freedom to add additional values in the future. Just having them documented is nice..

djmitche: Sure, but I have no idea what the acceptable values are -- and missing one
will cause 500 errors when an instance is in that state..
​

jonasfj:
Generally I see nothing wrong with this... However, Eli recently game a lecture on how we should plural form in REST APIs.

I suspect something like this:

* `GET /worker-types/` (listWorkerTypes)
* `GET /worker-types/:workerType` (workerType)
* `PUT /worker-types/:workerType` (createWorkerType)
* `POST /worker-types/:workerType/update` (updateWorkerType)
* `DELETE /worker-types/:workerType` (removeWorkerType)
* `GET /worker-types/:workerType/capacity` (capacity)
* `GET /worker-types/:workerType/resources` (resources)
* `GET /worker-types/:workerType/launch-specifications` (getLaunchSpecs)


-----
Final note, now that we've identified `capacity` as data that can be public I would suggest we we modify the listWorkerTypes result to be a list of `capacity`-like responses...

If we have the state in-memory on the server, then it's not a lot of overhead to always send... But it will dramatically reduce the delay and flickering in the aws-provisioner tool. (assuming that we send the capacity state from an in-memory cached representation of the ec2-state).

djmitche:
Worth noting that I've added this as a new API method, but once tools is using that, I'd like to rename that (copy/paste, modify tools to use the new name, remove old).

src/api-v1.js
```diff
@@ -151,9 +151,83 @@ async function validateWorkerType (ctx, workerTypeName, workerType) {
...
+
+  return summary;
+}
+
 module.exports = api;
 
 api.declare({
+  method: 'get',
+  route: '/worker-types',
+  name: 'listWorkerTypeSummaries', // TODO will become listWorkerTypes after transition
```
jhford: For authenticated users, there's no reason to calculate the capacities and other status information, and there's no reason to force these calculations if we're only wanting to list the worker types.  This is a lot of extra work if it's not needed.

src/api-v1.js
```diff
@@ -151,9 +151,83 @@ async function validateWorkerType (ctx, workerTypeName, workerType) {
...
+  });
+
+  return summary;
+}
+
 module.exports = api;
 
 api.declare({
+  method: 'get',
+  route: '/worker-types',
```
jhford: please keep this consistent with the rest of the file.  `/list-worker-type-summaries`.

src/api-v1.js
```diff
@@ -151,9 +151,83 @@ async function validateWorkerType (ctx, workerTypeName, workerType) {
...
+    'current capacity for each.  While this list includes all defined worker types,',
+    'there may be running EC2 instances for deleted worker types that are not',
+    'included here.  The list is unordered.',
+  ].join('\n'),
+}, async function (req, res) {
+  try {
+    // gather workerType information
+    let workerTypes = [];
+    await this.WorkerType.scan({}, {
+      handler: (item) => workerTypes.push(item),
```
jhford: The state look up should go in this handler so that we can begin calculating the summary while we're scanning the table.

test-src/api_test.js
```diff
@@ -0,0 +1,174 @@
+let workerType = require('../lib/worker-type');
```
jhford: We already have a file that tests the api, it's called manageworkertype_test.js.  I don't think there's any reason for these tests to be in a different file.

test-src/api_test.js
```diff
@@ -0,0 +1,174 @@
...
+
+    wt = slugid.nice();
+  });
+
+  afterEach(async () => {
+    await cleanTables();
+  });
+
+  let addTestWorkerType = async () => {
+    await WorkerType.create(wt, {
```
jhford: We have a module for creating mock worker types.  Please use that instead so that when we change the schema or entity, we don't need to make changes all over the place.  `manageworkertype_test.js` has example implementations.

test-src/api_test.js
```diff
@@ -0,0 +1,174 @@
...
+        {instanceType: 'm2', capacity: 2},
+      ],
+      regions: [],
+      lastModified: new Date(),
+      userData: {},
+      launchSpec: {},
+      secrets: {},
+      scopes: [],
+    });
+    await WorkerState.create({
```
jhford: Instead of coding these into the test directly, can you please split them out like we have for worker type definitions?  it makes updating schemas and entities a whole lot easier.

src/api-v1.js
djmitche: We don't want to have different response types depending on whether a user is logged in or not.  And this is still helpful for authenticated users, otherwise they need to go get a coffee while the UI loads every workerType.

src/api-v1.js
djmitche: Actually, I'm making this consistent with other services and REST principals, so I'd prefer to keep `/worker-types`.  I could be convinced to use `/worker-type` since the rest of the endpoints are also singular, but as jonas suggested we may want to change those some day, too.

test-src/api_test.js
djmitche: OK -- that's not a very obvious name, so I didn't find it :)

src/api-v1.js
djmitche: That's a minor savings in time (the scan results are streamed in a single http transaction, so `handler` is essentially called in a `.forEach` loop with no delay between), but I'll give it a shot.

src/api-v1.js
djmitche: Oh, are you suggesting that we keep two list-worker-types methods?

djmitche: Meh, OK, not my circus, not my monkeys :)

src/api-v1.js
djmitche: Meh, circus, monkeys.

djmitche:
OK, updated.  Ready to merge?

djmitche:
By the way, I think the update to fetch worker state earlier is *far* less readable.  I'm happy to keep it in if you like it, but equally happy to remove that commit.

src/api-v1.js
jhford: Let's do that when we create `/v2/` and modify all routes to use the plurals.  These routes have been this way forever and are used in existing tools that have been written.  I don't want to break everything right now.

selenamarie:
@jhford can we merge this?

jhford:
yes!

jhford:
I was under the assumption that this code had been tested. It had not been because it did not pass the linter, did not pass the unit test suite and does not work once deployed. I've backed it out.

PR #68, opened 2016-04-14
Bug1203083

jhford:
This also contains the check-for-ami pull request, but that's a different PR and is already R+'d.  Only cf12fcb should be considered in this PR


jonasfj:
Typing with nose because both hands are in the air... This is awesome...!
`workerType.description` is going to make management of workerTypes so much easier. I won't have to ping people on IRC to find out what a worker is all about... :)

(You might want to view previous warning about bitrot in separate pr for amis exist)


# r+

PR #69, opened 2016-04-14
Bug 1261912 - Remove or document the backendStatus() method

jhford:


schemas/backend-status-response.yml
```diff
@@ -0,0 +1,22 @@
...
+  Backend Status Response
+type:                       object
+properties:
+  status:
+    type: string
+    description: |
+      A string from Deadman's Snitch which describes the status.  See
+      https://deadmanssnitch.com/docs/api/v1#listing-your-snitches for an
+      explanation of this value
+  checked_in_at:
```
jonasfj: Nit, but maybe don't use `_`, we don't do this anywhere...

I would suggest `updated` or something..

Regardless, you probably want `format: date-time`, then use `new Date().toJSON()` which is compatible.

jonasfj:
I think you can just call it `status()` and `/status`.

I would definitely consider `format: date-time`... see comment...

jonasfj:
otherwise, it looks sane...

jhford:
This was landed as 790a2044

PR #70, opened 2016-04-19
Bug 1261910: add workerType to state().summary

djmitche:
This just makes the summary match that returned from `listWorkerTypeSummaries()`.  It also would have caused an error when we added a schema to `state()`

I'll work around this for the moment in tools with `_.assign({workerType: workerType}, ..)`


jhford:
I was under the assumption that this code had been tested.  It had not been because it did not pass the linter, did not pass the unit test suite and does not work once deployed.  I've backed it out.

djmitche:
We really should set up Travis :(

PR #72, opened 2016-04-21
Bug 1198391 (try 2): share summaries of workerTypes without requiring scopes

djmitche:
This passes all API-related tests, and eslint, unlike #67


src/api-v1.js
```diff
@@ -152,9 +152,94 @@ async function validateWorkerType (ctx, workerTypeName, workerType) {
...
+    'current capacity for each.  While this list includes all defined worker types,',
+    'there may be running EC2 instances for deleted worker types that are not',
+    'included here.  The list is unordered.',
+  ].join('\n'),
+}, async function (req, res) {
+  try {
+    // gather workerType information by scanning, and immediately begin a request
+    // for the corresponding workerState for each worker type, collecting promises
+    // that will yield {workerType, workerState}, possibly with the latter omitted.
+    let workerInfoPromises = [];
```
jhford: my intention for moving the logic into the handler was to do the work inside the handler itself.  If we're just going to build the list of promises then .all() them outside, the original way was easier to read.  My intention was to have the await happen inside the handler.

test-src/manageworkertype_test.js
```diff
@@ -57,3 +59,237 @@ describe('provisioner worker type api', () => {
...
+      credentials: helper.cfg.get('azure'),
+      context: {keyPrefix, provisionerId, provisionerBaseUrl, pubKey},
+    });
+
+    await WorkerState.ensureTable();
+    await WorkerType.ensureTable();
+    await cleanTables();
+
+    wt = slugid.nice();
+    testWorkerType = {
```
jhford: this still isn't using the mock-workers.js file.  Using that file means that when we do schema changes, we only need to update that single file to catch almost all instances of the update to the schema.

test-src/manageworkertype_test.js
```diff
@@ -57,3 +59,237 @@ describe('provisioner worker type api', () => {
...
+        {instanceType: 'm2', capacity: 2},
+      ],
+      regions: [],
+      lastModified: new Date(),
+      userData: {},
+      launchSpec: {},
+      secrets: {},
+      scopes: [],
+    };
+    testWorkerState = {
```
jhford: could we move this into a new file that works like mock-workers.js but called mock-workerstate.js?  It makes it easier to reuse this in other tests and makes updating schemas a lot easier

jhford:
Looks good!  If you can make the changes above, I'm happy to merge this and do a deployment.

test-src/manageworkertype_test.js
djmitche: I put it in `mock-workers.js`, since the WorkerType and WorkerState need to correspond.

djmitche:
OK, updated

jhford:
Looks good to me.  I'm going to run unit tests and merge if it all goes well :)

PR #73, opened 2016-04-21
use the upstream promise wrapper instead of ours

jhford:
dep left in because other things need it


jhford:
oh, this is not complete, will need to stop using the .data property

jhford:
agh, this is more complicated than I thought... i'll do this later...

PR #75, opened 2016-04-27
store the internally tracked state inside WorkerState entities

jhford:


src/aws-manager.js
```diff
@@ -1370,6 +1370,20 @@ class AwsManager {
       }
     }
 
+    for (let request of this.__internalState) {
+      if (request.WorkerType === workerName) {
+        response.internalTrackedRequests.push({
```
jonasfj: Two small things... Can we rename the variable to something other than response? To avoid perpetuating legacy...
Also I think you wanted to delete the to-do below...

jonasfj:
# `r+`

jhford:
this will also close #74

PR #76, opened 2016-04-28
clean up docs, get rid of state emulation function

jhford:


src/aws-manager.js
```diff
@@ -31,18 +89,32 @@ class AwsManager {
...
+
+    // We store the state of the previous iteration so that we can do
+    // comparisons to see which instances have changed state
     this.__previousApiState = this.__apiState;
+
+    // Internal state is state which we know exists but has not yet been
+    // reflected in the EC2 api.  This list is used to track the spot requests
+    // which we have submitted but which do not yet show up in the ec2 api
+    // calls which list the spot requests
     this.__internalState = [];
```
jonasfj: I really wish this was called something more intuitive... Like `_submittedSpotRequests` though that is way too long... :)
Just saying...

jonasfj: But if you decide to refactor that someday... maybe just make one list of everything spot request in any state, instances in any state, etc... And then a single place documenting what the shape of the objects is.

jonasfj:
from a quick look this seem pretty harmless, r+

src/aws-manager.js
jonasfj: ^ this is not a request to changes... just a random comment/observation.

jhford: yah.  there will be a refactor, and when it happens the changes will be large and far reaching.  this name likely won't survive.  `_submittedSpotRequests` isn't really right, since that's all spot requests, not just these ones.  The most specific name would be `_submittedButNotYetVisibleToApiSpotRequests` and I think that's for sure way too long!

We already have one list of all live instances and all live spot requests.  The dead instances and requests aren't really all that important outside of figuring out termination reasons, so no need to mess up every single other function with and extra `.filter` call, imo.

PR #77, opened 2016-05-09
Restrict regions to allowedRegions for creating and updating worker type definitions

petemoore:
I wasn't sure if there was a way to specify the values from https://github.com/taskcluster/aws-provisioner/blob/1ff88cf0f58083afd1b85904d6f57ead4eda634c/config/production.js#L11 directly here, rather than the way I've done it here (i.e. hard coding). If there is a way that we can reference the config value here, that might be nicer, so that if `allowedRegions` gets updated, or provisioner is deployed somewhere with different values, they will get picked up in the generated schemas.

The idea of this change is: a) documenting that only some regions are supported and b) ensure that only valid regions are accepted via the json interface.

### DIFF
--- a/schemas/create-worker-type-request.yml
+++ b/schemas/create-worker-type-request.yml
@@ -141,6 +141,10 @@ properties:
           type: string
           description: |
             The Amazon AWS Region being configured.  Example: us-west-1
+          enum:
+            - us-west-2
+            - us-east-1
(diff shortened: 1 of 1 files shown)


jhford:
I'd almost prefer to do this programmatically in the API by reading the live configuration value, but this works.  let's do this.

PR #78, opened 2016-05-10
bug 1217088 -- stop looking up scopes from worker type definition

jhford:
Basically ignore the scopes list from the worker type definition.  This list
should be removed later on

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -671,7 +671,7 @@ api.declare({
       credentials: taskcluster.createTemporaryCredentials({
         scopes: [
           `assume:worker-type:${this.provisionerId}/${secret.workerType}`,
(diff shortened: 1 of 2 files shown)


djmitche:
@jhford I think Jonas is away, but garndt r+'d in bug, and this looks good to me, too.  Do you want to land today?

src/api-v1.js
```diff
@@ -669,7 +669,10 @@ api.declare({
       data: secret.secrets,
       scopes: secret.scopes,
```
jonasfj: don't you also need to remove this?

jonasfj: @jhford, nvm, I assume you've planned some cleanup work to remove the workerType.scopes property later...

PR #79, opened 2016-05-10
Better error messages!

jhford:


petemoore:
This looks great!

Will we still get a 500 status code though if a bad AMI is specified?

jhford:
yep!  but now with an error message that's not likely the opposite of the real error case :)

PR #80, opened 2016-05-10
This would give each worker a unique id and matching scope

jhford:
This is instead of using instance id or spot request id.

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -333,14 +333,11 @@ api.declare({
       'launchSpec',
       'userData',
       'secrets',
-      'scopes',
       'minCapacity',
       'maxCapacity',
       'scalingRatio',
       'minPrice',
       'maxPrice',
-      'canUseOndemand',
-      'canUseSpot',
       'instanceTypes',
       'regions',
     ].every((key) => {
@@ -367,8 +364,6 @@ api.declare({
   // anymore, but left in because it's harmless and there might still be an
   // entity or two that require it
   let workerjson = wType.json();
-  delete workerjson.canUseOnDemand;
-  workerjson.canUseOndemand = false;
   res.reply(workerjson);
   return;
 });
@@ -484,8 +479,6 @@ api.declare({
     // We do this because John made a mistake in the V1->V2
     // schema update and there was a typo :(
     let workerjson = worker.json();
-    workerjson.canUseOndemand = false;
-    delete workerjson.canUseOnDemand;
 
     return res.reply(workerjson);
   } catch (err) {
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -6,6 +6,7 @@ let _ = require('lodash');
 let rp = require('request-promise');
 let url = require('url');
 let assert = require('assert');
+let slugid = require('slugid');
 
 let SLUGID_PATTERN = /^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$/;
 let GENERIC_ID_PATTERN = /^[a-zA-Z0-9-_]{1,22}$/;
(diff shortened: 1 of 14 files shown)


src/api-v1.js
```diff
@@ -658,13 +659,17 @@ api.declare({
       token: token,
     });
 
+    let workerId = slugid.v4();
+    let workerGroup = 'ec2_' + secret.region;
```
djmitche: I think the workerGroup should be `aws-provisioner` (maybe with `-v1`).  Without the instance ID, the region doesn't make a lot of sense.

djmitche:
This seems to be mixed in with a bunch of other refactors that should be broken out into their own commits, if not a different PR.  But the bits about slugIds work for me.

It's worth noting that releng is currently discussing "auditing" the instances on which tasks ran before making a release.  I'm not sure what that would mean or entail, and I bet it's just as easy with provisioner-specific instance identifiers as with AWS identifiers.

jonasfj:
note, that workerIds only have to be unique within the given workerGroup.

I'm curious what is the motivation...

I think it makes sense to use:
```
workerId = "<instanceId>";
workerGroup = "ec-<region>-<availability-zone>";
```
Then aws-provisioner-v1 just gets the scope: "queue:worker-id:ec2-*".


I think think it really matters... And we probably shouldn't build too many things that relies on this pattern. But when looking at a task it's convenient to easily have the instanceId and region.

So I'm curious what is the motivation of changing this? EC2 does reuse instanceIds, but usually not at the same time :)
Note: I don't really opppose this... Just curious why?

jhford:
@djmitche yes, there are a couple of PRs that I wrote this on top and forgot I had done so until after I had already pushed the PR :/  Sorry!

@jonasfj yes, it's handy, but I'd say that the place for that information is somewhere other than the scopes, as long as there are real reasons to do not put it in the scope.  Did you see the email that I wrote to accompany this PR?  EC2 make no guarantees about reuse of the id that I know other than that they can be reused.

We're also not strictly changing from instance ids to uuids, we're changing from a star scope for all worker groups and all instances to a specific per-instance value.

jhford:
I don't know how the worker-id scope is used on the docker worker.  @gregarndt maybe you can speak to this?

For deployment, I was thinking we could do a staged deployment:

1. set the workerId & workerGroup value in the return of the secret value, but still return the creds with star scope
2. update docker-worker to use the value from the secret to perform its operations against the queue
3. set the scope in the credential to be the resolved group and id value

gregarndt:
Doing a quick search within docker-worker it seems that it's set once during config loading, right now either specified through config or pulling from aws metadata, and then that value is used throughout the worker so I don't see any issue once we have the worker ID from somewhere.

As long as we record somewhere the instance ID from AWS for debugging purposes, I think things will be fine.

djmitche:
The reason not to use instanceId as workerId is that we are not guaranteed to ever get an instanceId from the AWS API (by a strict reading of "eventually consistent", and @jhford  and @rail have both seen the API not have an instanceId even after the instance has started).

I have frequently copy-pasted the instanceId out of a task inspector and searched for that in papertrail in order to find logs from that instance.  Switching to random identifiers would prevent that.

I talked to @escapewindow about auditing release graphs.  For forensic and auditing purposes, releng would like to be able to retrospectively determine "where" things ran.  For auditing purposes, a random ID is probably fine, as it could be correlated with other tasks run on on that same workerId.  For forensics, we would want an instance Id to reference AWS and papertrail logs, but that would be afterward.  So, yes, as long as there's some record eventually clearly linking a random identifier to an instanceId, I think this change would be OK.

jonasfj:
> The reason not to use instanceId as workerId is that we are not guaranteed to ever get an instanceId from the AWS API

I buy that argument, and is at the same time very sad...

Note: workerGroup/workerId is used when claiming tasks. It ensures that runs are somehow linked to the worker they ran on. The worker only cares in the sense that it must send them to the queue..
That said it might make sense for the worker to tag sentry submissions with workerGroup/workerId, but that's a different story.

jhford:
Closed in favour of an patch which we'd consider merging (https://github.com/taskcluster/aws-provisioner/pull/81) for bug 1272643

PR #81, opened 2016-05-13
bug 1272643 - start providing workerGroup and workerId value to workers

jhford:
The only open question here is whether to make the workerGroup be the provisionerId or e.g. ec2_us-west-1.  I don't have a strong preference one way or the other, but including the region is 90% of the content of this patch, as it requires an update to the secret entity.

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -6,6 +6,7 @@ let _ = require('lodash');
 let rp = require('request-promise');
 let url = require('url');
 let assert = require('assert');
+let slugid = require('slugid');
 
 let SLUGID_PATTERN = /^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$/;
 let GENERIC_ID_PATTERN = /^[a-zA-Z0-9-_]{1,22}$/;
@@ -603,7 +604,7 @@ api.declare({
       token: token,
       workerType: input.workerType,
       secrets: input.secrets,
-      scopes: input.scopes,
+      region: input.region,
       expiration: new Date(input.expiration),
     });
   } catch (err) {
@@ -620,7 +621,7 @@ api.declare({
       'workerType',
       'secrets',
       'token',
-      'scopes',
+      'region',
       //'expiration' weird stuff is happening here.  going to assume that
       // we should probably do some sort of Date.toISOString() comparison or something
     ].every((key) => {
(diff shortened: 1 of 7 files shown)


jhford:
Greg, what do you think about this change?  This will still issue the * scope, but will tell the worker what it *should* be using for workerId and group

src/api-v1.js
```diff
@@ -665,13 +666,17 @@ api.declare({
       token: token,
     });
 
+    let workerId = slugid.v4();
+    let workerGroup = 'ec2_' + secret.region;
```
jonasfj: Bike shedding here but with dash it gets hard to parse the workerGroup... So people are less likely to use it in any form of automation... Also it's prettier :)

Regardless this.Secret.load({token}) also works....


Oh, and we need to at-least write the mapping between spotreqid or something and the workerId... Or maybe log requested IP too... Console.log is better than nothing... Queue has a module to get IP from request...

Also what if there is two requests here... There can be due to retires... Do we want to issue the same workerId?
I'm thinking we don't... But maybe we need to spend two braincycles considering the implications...

jonasfj: Actually maybe it's better we don't do that in this PR... Probably we should limit number of retires allowed per token...

jonasfj:
What * scopes?

jonasfj:
On topic: why make workgroup = provisionerId? It already given elsewhere...
Having a regions seems nice... As long as it's prefix ec2 :)

src/api-v1.js
```diff
@@ -665,13 +666,17 @@ api.declare({
...
     return res.reply({
       data: secret.secrets,
-      scopes: secret.scopes,
+      workerId: workerId,
+      workerGroup: workerGroup,
       credentials: taskcluster.createTemporaryCredentials({
         scopes: [
           `assume:worker-type:${this.provisionerId}/${secret.workerType}`,
-          'assume:worker-id:*',
+          `assume:worker-id:${workerGroup}:${workerId}`,
```
jonasfj: I think this is a legacy scope...

See queue sources it's in the comments of the API definition...
The new line is queue:worker-id:... If I recall correctly..

jonasfj: I know it's bad but I can't remove the old one till people stops using it...

src/api-v1.js
```diff
@@ -665,13 +666,17 @@ api.declare({
       token: token,
     });
 
+    let workerId = slugid.v4();
```
gregarndt: Is there any length limit on the worker ID from our standpoint since it'll be included in scopes?

src/api-v1.js
gregarndt: So it looks like we're changing workerGroup to be ec2_<region> instead of just <region> which is ok as far as I'm concerned.  I would like us to maintain region somehow because we use that in metrics and logs for knowing which region a particular set of instances came from.

jonasfj: 22 chars... Has to fit in routing key

gregarndt:
This has sat opened for some time now and has merge conflicts too so probably bitrotted a bit.  Do we want to close this for now or should we proceed with trying to implement this?

src/api-v1.js
jonasfj: Yes, it's in pulse routing keys too..

jonasfj:
Yes, we should be issuing `queue:worker-id:<workerGroup>/<workerId>`

djmitche:
Sounds like we agreed to close this 2 years ago :)

PR #82, opened 2016-05-26
Provisioner updates

jhford:
I got so sick of remembering how the old boilerplate and config stuff worked so I decided to bring the provisioner up to date.  This is the first step in changing which azure account we use.


jonasfj:
YAY!!!!

src/main.js
```diff
@@ -0,0 +1,275 @@
...
+    },
+  },
+
+  validator: {
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      return await base.validator({
+        folder: path.join(__dirname, '..', 'schemas'),
+        constants: require('../schemas/constants'),
+        publish: cfg.app.publishMetaData === 'true',
```
jonasfj: newest validator guess this using `NODE_ENV == 'production'`... I think...

@imbstack, can you help review this main.js, specifically, you are 100% sure about what is best practice and not...

src/main.js
```diff
@@ -0,0 +1,275 @@
...
+  provisioner: {
+    requires: ['cfg', 'awsManager', 'WorkerType', 'WorkerState', 'Secret', 'ec2', 'influx'],
+    setup: async ({cfg, awsManager, WorkerType, WorkerState, Secret, ec2, influx}) => {
+      let queue = new taskcluster.Queue({credentials: cfg.taskcluster.credentials});
+
+      let provisioner = new provision.Provisioner({
+        WorkerType: WorkerType,
+        Secret: Secret,
+        WorkerState: WorkerState,
+        queue: queue,
```
jonasfj: It's probably easier to write and read... if you just do: {WorkerType, Secret, WorkerState, queue, provisionerId: cfg.app.id, ....

but also a nit...

src/main.js
```diff
@@ -0,0 +1,275 @@
...
+        awsManager: awsManager,
+        provisionIterationInterval: cfg.app.iterationInterval,
+        dmsApiKey: cfg.deadmanssnitch.api.key,
+        iterationSnitch: cfg.deadmanssnitch.iterationSnitch,
+      });
+
+      try {
+        provisioner.run();
+      } catch (err) {
+        debug('[alert-operator] Error: ' + err.stack || err);
```
jonasfj: monitor.reportError(err) will make your life so much better :)

jhford: yup!  need to figure out the stats stuff as it relates to the biaser before I rip out lib-stats and put in lib-monitor.

config.yml
```diff
@@ -0,0 +1,99 @@
+defaults:
+  app:
+    publishMetaData: false
```
jonasfj: `!env:bool PUBLISH_METADATA` is what I would use...

Maybe compare to: https://github.com/taskcluster/taskcluster-queue/blob/master/config.yml

jonasfj: it defaults to false...

config.yml
```diff
@@ -0,0 +1,99 @@
...
+    publicUrl: https://localhost:60407
+    port: !env:number PORT
+    env: test
+    forceSSL: false
+    trustProxy: false
+  deadmanssnitch:
+    api:
+      key: 'nokey'
+    iterationSnitch: 'http://httpbin.org/status/200'
+
```
jonasfj: In `test` please set the `azure.account = "jungle"` this is the garbage account...
You test environment shouldn't have scopes to talk to the production account...

config.yml
```diff
@@ -0,0 +1,99 @@
...
+    iterationInterval: 10000
+    maxInstanceLife: '- 96 hours'
+    id: aws-provisioner-dev
+    workerTypeTableName: DevWorkerTypes
+    secretTableName: DevWorkerSecrets
+    workerStateTableName: DevWorkerState
+    awsInstancePubkey: !env AWS_INSTANCE_PUBKEY
+    awsKeyPrefix: "aws-provisioner-dev-managed:"
+    statsComponent: aws-provisioner-dev
+    allowedRegions: "us-west-2,us-east-1,us-west-1"
```
jonasfj: `!env:list ALLOWED_REGIONS`

Then configure with ALLOWED_REGIONS="us-west-2 us-east-1 us-west-1"
If you prefer to hard code it,... then just do that in the `production` section...
`allowedRegions: ['us-west-2', 'us-east-1', 'us-west-1']` or even use YAML-style lists :)

jonasfj: This way you don't have to do any string splitting in-tree...

user-config-example.yml
```diff
@@ -0,0 +1,11 @@
+defaults:
+  aws:
+    accessKeyId:        '...'
+    secretAccessKey:    '...'
+    region:             us-west-2
+  azure:
+    accountName:        jungle
```
jonasfj: YAY, remove the `accountKey` entry... no need... you can easily get jungle access using tc credentials..

jonasfj:
# `r+`

I think my comments are minor... from skimming it generally seems okay... Definitely better than what we had before...

src/main.js
jhford: i guess I'm weird, because I really find that syntax way less clear.

jonasfj: okay, that's fine... I can totally live with this...

Personally, I find this without alignment, hard to read... on the plus side I do see that it forces me to read it more closely :)

Anyways, it's not remotely important... I just wanted to highlight that you had options...

PR #84, opened 2016-05-30
switch to azure blob storage for worker state entries

jhford:
pretty small change in the end.  I kind of like the container thing I wrote here, maybe we could put it in taskcluster base and make it a little more general.


jonasfj:
# `r+`

jonasfj:
I did at one point start adding blob support to fast-azure-storage but never finished it... there is a lot of features in blob storage...

PR #85, opened 2016-06-06
Add AMI set entity and its create and delete endpoints

anarute:
I was not sure about its endpoints, if I am using the correct AMI Set properties.

### DIFF
--- a/schemas/create-ami-set-request.yml
+++ b/schemas/create-ami-set-request.yml
@@ -1,0 +1,33 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      Create AMI Set Request
+description: |
+  Create a new AMI Set
+type:                       object
+properties:
+  amis:
+    type: array
+    description: |
+      AMIs in this set
+    items:
+      type: object
+      properties:
+        region:
+          type: string
+          description: |
+            The Amazon AWS region of the AMI set
+        hvm:
+          type: string
+          description: |
+            The AMI that uses HVM virtualization type
+        pv:
+          type: string
+          description: |
+            The AMI that uses PV virtualization type
+      additionalProperties: false
+      required:
+        - region
+        - hvm
+        - pv
+additionalProperties: false
+required:
+  - amis
--- a/schemas/get-ami-set-response.yml
+++ b/schemas/get-ami-set-response.yml
@@ -1,0 +1,44 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Get AMI Set Response"
+description: |
+  An AMI Set from the provisioner
+type:                       object
+properties:
+  id:
+    type: string
+    description: |
+      Name of the AMI set
+  amis:
+    type: array
+    description: |
+      AMIs in this set
+    items:
+      type: object
+      properties:
+        region:
+          type: string
+          description: |
+            The Amazon AWS region of the AMI set
+        hvm:
+          type: string
+          description: |
+            The AMI that uses HVM virtualization type
+        pv:
+          type: string
+          description: |
+            The AMI that uses PV virtualization type
+      additionalProperties: false
+      required:
+        - region
+        - hvm
+        - pv
+  lastModified:
+    type: string
+    format: date-time
+    description: |
+      ISO Date string (e.g. new Date().toISOString()) which represents the time
+      when this AMI Set was last altered (inclusive of creation)
+additionalProperties: false
+required:
+  - id
+  - amis
--- a/src/ami-set.js
+++ b/src/ami-set.js
@@ -1,0 +1,73 @@
+let base = require('taskcluster-base');
+let debug = require('debug')('aws-provisioner:AmiSet');
+
+const KEY_CONST = 'ami-set';
+
+ /**
+ * An AMI Set is a collection of AMIs with a single name (its AMI Set ID).
+ * Each AMI in the set is keyed by its virtualization type (PV or HVM) and
+ * by its AWS region.
+ */
+
+let AmiSet = base.Entity.configure({
+  version: 1,
+
+  partitionKey: base.Entity.keys.ConstantKey(KEY_CONST),
+  rowKey: base.Entity.keys.StringKey('id'),
+
+  properties: {
+
+    id: base.Entity.types.String,
+    /* This is a JSON object which contains the AMIs of an AMI set keyed by
+     * their virtualization type and region. It is in the shape:
+     * [
+     *   {
+     *     region: us-west-1,
+     *     hvm: ami-2222,
+     *     pv: ami-3333
+     *   }
+     * ]
+     */
+    amis: base.Entity.types.JSON,
+    // Store the date of last modification for this entity
+    lastModified: base.Entity.types.Date,
+
+  },
+});
+
+/**
+ * Load all the knwon Amis
+ */
+AmiSet.listAmiSets = async function () {
+
+  let amiSetList = [];
+
+  try {
+    await base.Entity.scan.call(this, {}, {
+      handler: function(item) {
+        amiSetList.push(item.json());
+      },
+    });
+  } catch (err) {
+    debug('error listing Ami sets');
+    debug(err);
+    if (err.stack) {
+      debug(err.stack);
+    }
+    throw err;
+  }
+
+  return amiSetList;
+};
+
+/**
+ * Return an Object for JSON encoding which represents
+ * the data associated with this AMI Set.  This is a
+ * method intended for use in displaying the data associated
+ * with a given amiSet
+ */
+AmiSet.prototype.json = function() {
+  return JSON.parse(JSON.stringify(this._properties));
+};
+
+module.exports = AmiSet;
--- a/src/ami-set.js
+++ b/src/ami-set.js
@@ -36,16 +36,15 @@ let AmiSet = base.Entity.configure({
 });
 
 /**
- * Load all the knwon Amis
+ * Load the names of all the known Amis
  */
 AmiSet.listAmiSets = async function () {
-
-  let amiSetList = [];
+  let names = [];
 
   try {
     await base.Entity.scan.call(this, {}, {
       handler: function(item) {
-        amiSetList.push(item.json());
+        names.push(item.id);
       },
     });
   } catch (err) {
@@ -57,7 +56,7 @@ AmiSet.listAmiSets = async function () {
     throw err;
   }
 
-  return amiSetList;
+  return names;
 };
 
 /**
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -125,7 +125,7 @@ async function validateWorkerType(ctx, workerTypeName, workerType) {
           ImageIds: [launchSpec.ImageId],
         }).promise();
         if (images.data.Images.length !== 1) {
-          reasons.push(new Error(`Too many results found for ${launchSpec.ImageId}`)); 
+          reasons.push(new Error(`Too many results found for ${launchSpec.ImageId}`));
         }
         let image = images.data.Images[0];
         if (image.ImageId !== launchSpec.ImageId) {
@@ -144,7 +144,7 @@ async function validateWorkerType(ctx, workerTypeName, workerType) {
   // Finally, let's verify that the image for this workerType exists in EC2
 
   if (reasons.length > 0) {
-    debug('Found errors with ' + workerTypeName); 
+    debug('Found errors with ' + workerTypeName);
     let e = new Error('Refusing to create an invalid worker type');
     e.reasons = reasons;
     throw e;
@@ -575,6 +575,183 @@ api.declare({
       debug(err.stack);
     }
     throw err;
+  }
+});
+
+api.declare({
+  method: 'put',
+  route: '/ami-set/:id',
+  name: 'createAmiSet',
+  deferAuth: true,
+  scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  title: 'Create new AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let id = req.params.id;
+
+  // Authenticate request with parameterized scope
+  if (!req.satisfies({id})) {
+    return;
+  }
+
+  // Create amiSet
+  let amiSet;
+  try {
+    amiSet = await this.AmiSet.create(id, input);
+  } catch (err) {
+    // We only catch EntityAlreadyExists errors
+    if (!err || err.code !== 'EntityAlreadyExists') {
+      throw err;
+    }
+    amiSet = await this.AmiSet.load({id});
+
+    // Check if it matches the existing amiSet
+    let match = [
+      'amis',
+    ].every((key) => {
+      return _.isEqual(amiSet[key], input[key]);
+    });
+
+    // If we don't have a match we return 409, otherwise we continue as this is
+    // is an idempotent operation.
+    if (!match) {
+      res.status(409).json({
+        error: 'AMI Set already exists with different definition',
+      });
+      return;
+    }
+  }
+
+});
+
+api.declare({
+  method: 'get',
+  route: '/ami-set/:id',
+  name: 'amiSet',
+  output: 'get-ami-set-response.json#',
+  deferAuth: true,
+  title: 'Get AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Retreive a copy of the requested AMI set.',
+  ].join('\n'),
+}, async function (req, res) {
+  let id = req.params.id;
+
+  let amiSet;
+  try {
+    amiSet = await this.AmiSet.load({id});
+    res.reply(amiSet.json());
+  } catch (err) {
+    if (err.code === 'ResourceNotFound') {
+      res.status(404).json({
+        error: err.code,
+        msg: id + ' not found',
+      });
+    } else {
+      throw err;
+    }
+  }
+});
+
+api.declare({
+  method: 'post',
+  route: '/ami-set/:id/update',
+  name: 'updateAmiSet',
+  deferAuth: true,
+  scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  title: 'Update AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Provide a new copy of an AMI Set to replace the existing one.',
+    'This will overwrite the existing AMI Set if there',
+    'is already an AMI Set of that name. This method will return a',
+    '200 response along with a copy of the AMI Set created.',
+    'Note that if you are using the result of a GET on the ami-set',
+    'end point that you will need to delete the lastModified and amiSet',
+    'keys from the object returned, since those fields are not allowed',
+    'the request body for this method.',
+    '',
+    'Otherwise, all input requirements and actions are the same as the',
+    'create method.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let id = req.params.id;
+
+  if (!req.satisfies({id})) {
+    return;
+  }
+
+  let amiSet = await this.AmiSet.load({id});
+
+  await amiSet.modify(function(amiSet) {
+    // We know that data that gets to here is valid per-schema
+    amiSet.amis = input.amis;
+  });
+  return res.reply(amiSet.json());
+});
+
+api.declare({
+  method: 'get',
+  route: '/list-ami-sets',
+  name: 'listAmiSets',
+  title: 'List AMI sets',
+  stability:  base.API.stability.stable,
+  description: [
+    'Return a list of AMI sets names.',
+  ].join('\n'),
+}, async function (req, res) {
+
+  try {
+    let list = await this.AmiSet.listAmiSets();
+    return res.reply(list);
+  } catch (err) {
+    debug('error listing amiSets');
+    debug(err);
+    if (err.stack) {
+      debug(err.stack);
+    }
+    throw err;
+  }
+});
+
+api.declare({
+  method: 'delete',
+  route: '/ami-set/:id',
+  name: 'removeAmiSet',
+  deferAuth: true,
+  scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  title: 'Delete AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Delete an AMI Set.',
+  ].join('\n'),
+}, async function (req, res) {
+  let id = req.params.id;
+
+  if (!req.satisfies({id})) {
+    return;
+  }
+
+  try {
+    await this.AmiSet.remove({id});
+    res.status(204).end();
+  } catch (err) {
+    if (err.code === 'ResourceNotFound') {
+      res.status(204).end();
+    } else {
+      debug('unknown error deleting AMI Set ' + id);
+      debug(err);
+      if (err.stack) {
+        debug(err.stack);
+      }
+      throw err;
+    }
   }
 });
 
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -582,6 +582,7 @@ api.declare({
   method: 'put',
   route: '/ami-set/:id',
   name: 'createAmiSet',
+  input: 'create-ami-set-request.json#',
   deferAuth: true,
   scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
   title: 'Create new AMI Set',
@@ -601,7 +602,11 @@ api.declare({
   // Create amiSet
   let amiSet;
   try {
-    amiSet = await this.AmiSet.create(id, input);
+    amiSet = await this.AmiSet.create({
+      id: id,
+      amis: input.amis,
+      lastModified: new Date(),
+    });
   } catch (err) {
     // We only catch EntityAlreadyExists errors
     if (!err || err.code !== 'EntityAlreadyExists') {
@@ -625,6 +630,8 @@ api.declare({
       return;
     }
   }
+  res.reply({outcome: 'success'});
+  return;
 
 });
 
@@ -664,6 +671,8 @@ api.declare({
   name: 'updateAmiSet',
   deferAuth: true,
   scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  input: 'create-ami-set-request.json#',
+  output: 'get-ami-set-response.json#',
   title: 'Update AMI Set',
   stability:  base.API.stability.stable,
   description: [
@@ -700,6 +709,7 @@ api.declare({
   method: 'get',
   route: '/list-ami-sets',
   name: 'listAmiSets',
+  output: 'list-ami-sets-response.json#',
   title: 'List AMI sets',
   stability:  base.API.stability.stable,
   description: [
@@ -726,6 +736,8 @@ api.declare({
   name: 'removeAmiSet',
   deferAuth: true,
   scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  input: undefined,  // No input
+  output: undefined,  // No output
   title: 'Delete AMI Set',
   stability:  base.API.stability.stable,
   description: [
--- a/src/main.js
+++ b/src/main.js
@@ -11,6 +11,7 @@ let base = require('taskcluster-base');
 
 let workerType = require('./worker-type');
 let secret = require('./secret');
+let amiSet = require('./ami-set');
 let AwsManager = require('./aws-manager');
 let provision = require('./provision');
 let exchanges = require('./exchanges');
@@ -56,6 +57,18 @@ let load = base.loader({
         },
       });
       return WorkerType;
+    },
+  },
+
+  AmiSet: {
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      let AmiSet = amiSet.setup({
+        account: cfg.azure.account,
+        table: cfg.app.amiSetTableName,
+        credentials: cfg.taskcluster.credentials,
+      });
+      return AmiSet;
     },
   },
 
@@ -124,13 +137,15 @@ let load = base.loader({
   },
 
   api: {
-    requires: ['cfg', 'WorkerType', 'Secret', 'ec2', 'stateContainer', 'validator', 'publisher', 'influx'],
-    setup: async ({cfg, WorkerType, Secret, ec2, stateContainer, validator, publisher, influx}) => {
+    requires: ['cfg', 'WorkerType', 'AmiSet', 'Secret', 'ec2', 'stateContainer', 'validator', 'publisher', 'influx'],
+    setup: async ({cfg, WorkerType, AmiSet, Secret, ec2, stateContainer, validator, publisher, influx}) => {
+
       let reportInstanceStarted = series.instanceStarted.reporter(influx);
 
       let router = await v1.setup({
         context: {
           WorkerType: WorkerType,
+          AmiSet: AmiSet,
           Secret: Secret,
           publisher: publisher,
           keyPrefix: cfg.app.awsKeyPrefix,
@@ -154,7 +169,7 @@ let load = base.loader({
         drain: influx,
       });
 
-      return router; 
+      return router;
     },
   },
 
@@ -175,13 +190,16 @@ let load = base.loader({
 
   // Table Cleaner for testing
   tableCleaner: {
-    requires: ['WorkerType', 'Secret'],
-    setup: async ({WorkerType, Secret}) => {
+    requires: ['WorkerType', 'Secret', 'AmiSet'],
+    setup: async ({WorkerType, Secret, AmiSet}) => {
       await Promise.all([
         WorkerType.scan({}, {
           handler: async (x) => { await x.remove(); },
         }),
         Secret.scan({}, {
+          handler: async (x) => { await x.remove(); },
+        }),
+        AmiSet.scan({}, {
           handler: async (x) => { await x.remove(); },
         }),
       ]);
@@ -247,7 +265,7 @@ let load = base.loader({
       } catch (err) {
         debug('[alert-operator] Error: ' + err.stack || err);
       }
-      
+
       return provisioner;
     },
   },
--- a/test-src/manageamiset_test.js
+++ b/test-src/manageamiset_test.js
@@ -1,0 +1,58 @@
+var slugid = require('slugid');
+var assume = require('assume');
+var debug = require('debug')('test');
+var _ = require('lodash');
+var mock = require('./mock-amisets');
+var base = require('taskcluster-base');
+
+var makeAmiSet = mock.makeAmiSet;
+
+var main = require('../lib/main');
+var helper = require('./helper');
+
+describe('AMI Set api', () => {
+
+  var id = slugid.nice();
+  var amiSetDefinition = makeAmiSet();
+  var amiSetChanged = _.clone(amiSetDefinition);
+
+  let AmiSet;
+
+  let client;
+
+  let testAmiSet = makeAmiSet({
+    lastModified: new Date(),
+  });
+
+  before(async () => {
+    AmiSet = await main('Amiset', {process: 'AmiSet', profile: 'test'});
+
+    client = helper.getClient();
+  });
+
+  beforeEach(async () => {
+    await main('tableCleaner', {process: 'tableCleaner', profile: 'test'});
+  });
+
+  it('should be able to create an AMI set (idempotent)', async () => {
+    debug('### Create AMI Set');
+    await client.createAmiSet(id, amiSetDefinition);
+
+    debug('### Create AMI Set (again)');
+    await client.createAmiSet(id, amiSetDefinition);
+  });
+
+  it('should be able to remove an AMI set (idempotent)', async () => {
+    debug('### Remove AMI Set');
+    await client.removeAmiSet(id);
+    await client.removeAmiSet(id);
+
+    debug('### Try to load AMI Set');
+    try {
+      await client.amiSet(id);
+      throw new Error('Expected and error');
+    } catch (err) {
+      assume(err.statusCode).equals(404);
+    }
+  });
+});
--- a/test-src/manageamiset_test.js
+++ b/test-src/manageamiset_test.js
@@ -12,20 +12,23 @@ var helper = require('./helper');
 
 describe('AMI Set api', () => {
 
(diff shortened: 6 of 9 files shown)


src/ami-set.js
```diff
@@ -0,0 +1,23 @@
...
+
+let amiSet = base.Entity.configure({
+  version: 1,
+
+  partitionKey: base.Entity.keys.ConstantKey(KEY_CONST),
+  rowKey: base.Entity.keys.StringKey('amiSetId'),
+
+  properties: {
+
+    amiSetId: base.Entity.types.String,
```
jhford: since we're in the AmiSet class, let's call this just 'id'

src/ami-set.js
```diff
@@ -0,0 +1,23 @@
...
+    amis: base.Entity.types.JSON,
```
jhford: Can you write a comment here that describes what this object should look like?  It's really helpful to know what to expect in the data when coding against this Entity in the future.

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
 
 api.declare({
   method: 'put',
+  route: '/ami-set/:amiSetId',
+  name: 'createAmiSet',
+  deferAuth: true,
+  scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  input: 'create-ami-set-request.json#',
+  output: 'get-ami-set-response.json#',
```
djmitche: These (input and output) point to the schemas we talked about today.  You can just omit these lines for the moment, and then return to add schemas in a later pull request.

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
...
+  title: 'Create new AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let amiSet = req.params.amiSetId;
+
+  input.lastModified = new Date();
```
djmitche: A `lastModified` property is a good idea, but you'll need to include it in `src/ami-set.js` as well.

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
...
+        error: 'AMI Set already exists with different definition',
+      });
+      return;
+    }
+  }
+
+  // Publish pulse message
+  await this.publisher.amiSetCreated({
+    amiSet: amiSetId,
+  });
```
djmitche: I don't think we need to send a pulse message on AMI set creation.  It doesn't hurt, and we could add it back in later.

For WorkerTypes, this message is required so that the provisioner backend can look at the new workerType and start creating EC2 instance for it.  For new AmiSets, there's no need for the backend to do anything.

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
...
+});
+
+api.declare({
+  method: 'delete',
+  route: '/ami-set/:amiSetId',
+  name: 'removeAmiSet',
+  deferAuth: true,
+  scopes: [['aws-provisioner:manage-ami-set:<amiSetId>']],
+  input: undefined,  // No input
+  output: undefined,  // No output
```
djmitche: You can just omit these

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
...
+  title: 'Delete AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Delete an AMI Set.',
+  ].join('\n'),
+}, async function (req, res) {
+  let that = this;
+  let amiSet = req.params.amiSetId;
+
+  if (!req.satisfies({amiSet: amiSetId})) { return undefined; }
```
djmitche: Why `undefined`?

src/main.js
```diff
@@ -11,6 +11,8 @@ let base = require('taskcluster-base');
 
 let workerType = require('./worker-type');
 let secret = require('./secret');
+let workerState = require('./worker-state');
```
jhford: You should rebase your patch to the latest version of the upstream master branch.  There was a change that I landed today that deletes the WorkerState entity, and it looks like you're adding it back here.

src/main.js
```diff
@@ -59,6 +61,30 @@ let load = base.loader({
...
+      let WorkerState = workerState.setup({
+        account: cfg.azure.account,
+        table: cfg.app.workerStateTableName,
+        credentials: cfg.taskcluster.credentials,
+      });
+      return WorkerState;
+    },
+  },
+
+  amiSet: {
```
jhford: We have a convention of anything that's a `taskcluster-base.Entity` using class casing, so please call this AmiSet here, and the other places that you are referring to the thing returned by `AmiSet.setup` instead of an instance of an `AmiSet`.

src/main.js
```diff
@@ -59,6 +61,30 @@ let load = base.loader({
...
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      let amiSet = amiSet.setup({
+        account: cfg.azure.account,
+        table: cfg.app.workerStateTableName,
```
djmitche: I think you'll need a different table name for ami sets :)

src/ami-set.js
```diff
@@ -0,0 +1,23 @@
...
+
+const KEY_CONST = 'ami-set';
+
+ /**
+ * An AMI Set is a collection of AMIs with a single name (its AMI Set ID).
+ * Each AMI in the set is keyed by its virtualization type (PV or HVM) and
+ * by its AWS region.
+ */
+
+let amiSet = base.Entity.configure({
```
djmitche: This should be capitalized (`AmiSet`).  The idea is that this is a class, so just like in Python we use an initial capital letter.

src/ami-set.js
```diff
@@ -0,0 +1,23 @@
...
+  partitionKey: base.Entity.keys.ConstantKey(KEY_CONST),
+  rowKey: base.Entity.keys.StringKey('amiSetId'),
+
+  properties: {
+
+    amiSetId: base.Entity.types.String,
+    amis: base.Entity.types.JSON,
+
+  },
+});
```
djmitche: The content of this module looks OK, but it doesn't export anything.  There's some information on exports here:

  https://www.sitepoint.com/understanding-module-exports-exports-node-js/

In this case, we want `require('./ami-set')` to return the configured entity, just like for `require('./worker-type')`.

src/api-v1.js
jhford: We don't actually consume any pulse messages in the provisioner.  We scan the worker type table on each iteration, since we need to load all the worker types anyway to see what the max/min capacity, scaling ratio and instance types.  We could maintain that state in the backend process, but that's not there yet and would be a fairly substantial code change.  I don't know if the Azure overhead (tech or billing wise) is worth the trouble.

That said, I don't think that we really need to have pulse messages for AmiSets.  The important Pulse messages are for WorkerType updates, since those are (well, should be) used to gracefully shutdown old instances when we update the worker type definition.

src/main.js
djmitche: This should be capitalized (`AmiSet`).

src/main.js
```diff
@@ -59,6 +61,30 @@ let load = base.loader({
...
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      let WorkerState = workerState.setup({
+        account: cfg.azure.account,
+        table: cfg.app.workerStateTableName,
+        credentials: cfg.taskcluster.credentials,
+      });
+      return WorkerState;
+    },
+  },
```
djmitche: I think this may be a merge error?  @jhford just removed the WorkerState entity class.  Awesome that you've rebased your changes against the latest master, by the way!

src/api-v1.js
```diff
@@ -580,6 +580,105 @@ api.declare({
...
+  input: 'create-ami-set-request.json#',
+  output: 'get-ami-set-response.json#',
+  title: 'Create new AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let amiSet = req.params.amiSetId;
```
jhford: I probably would call this `amiSetId` since you'll be creating an Entity and probably want this name for that, instead of `aSet`

src/main.js
```diff
@@ -124,13 +150,15 @@ let load = base.loader({
   },
 
   api: {
-    requires: ['cfg', 'WorkerType', 'Secret', 'ec2', 'stateContainer', 'validator', 'publisher', 'influx'],
-    setup: async ({cfg, WorkerType, Secret, ec2, stateContainer, validator, publisher, influx}) => {
+    requires: ['cfg', 'WorkerType', 'WorkerState', 'amiSet', 'Secret', 'ec2', 'validator', 'publisher', 'influx'],
+    setup: async ({cfg, WorkerType, WorkerState, amiSet, Secret, ec2, validator, publisher, influx}) => {
+
```
djmitche: As above, I don't think you want WorkerState in here (@jhford please correct me if I'm wrong!)

src/main.js
```diff
@@ -124,13 +150,15 @@ let load = base.loader({
...
       let reportInstanceStarted = series.instanceStarted.reporter(influx);
 
       let router = await v1.setup({
         context: {
           WorkerType: WorkerType,
+          amiSet: amiSet,
```
djmitche: This too should be capitalized (`AmiSet`)

src/main.js
jhford: Nope, she doesn't.  I think that a merge went a little wrong, because I just today deleted the WorkerState on the master branch.

jhford:
@anarute really good first pass!  Most of the things that are needed are here.  It seems like you and @djmitche are talking about json-schema for the AmiSets, which is a great idea.  Otherwise, looking great!

djmitche:
OK that was a lot of comments, but this looks good!

If I remember correctly, you were able to run the tests for the provisioner.  Do they still run after your changes?

You will also need some tests in `test-src/manage_ami_set_test.js`, similar to `test-src/manageworkertype_test.js` (but again, much shorter, since so far AMI sets are a lot less complicated than worker types).  If it would help you out, I can write up a test for one of the API methods as an example.

src/ami-set.js
anarute: oh, right! so, I think I also need an AmiSet.create function, right?

djmitche: You get that one for free -- `base.Entity.configure(..)` creates a class for you with a few methods (`create`, `update`, `remove`, etc.).  Sadly there's basically no documentation for that (https://bugzilla.mozilla.org/show_bug.cgi?id=1278385).

src/api-v1.js
anarute: ops, sorry, I think I misunderstood this from worker-type.js

djmitche: An understandable confusion -- `WorkerType.create` actually overrides the `Entity.create` method to give it a different function signature.  That's not necessary for AmiSet.

src/main.js
anarute: yes, it is a merge error! probably it was from the commit last week and I didn't see this part was removed. Now I see the importance of sending small patches as soon as I can :)

anarute:
Thank you @djmitche and @jhford for the comments and feedback! It was really helpful to understand things better. I *think* (if I have not added more things to fix) that I've covered most part of the issues, could you please review it again?

src/api-v1.js
```diff
@@ -584,41 +584,39 @@ api.declare({
...
-  output: 'get-ami-set-response.json#',
   title: 'Create new AMI Set',
   stability:  base.API.stability.stable,
   description: [
     'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
   ].join('\n'),
 }, async function (req, res) {
   let input = req.body;
-  let amiSet = req.params.amiSetId;
+  let AmiSetId = req.params.id;
```
jhford: AmiSet is a 'class', but this isn't referring to the class.  This is referring to a string and so should be amiSetId or just id... either one is fine.

src/api-v1.js
```diff
@@ -584,41 +584,39 @@ api.declare({
...
 
   // Authenticate request with parameterized scope
-  if (!req.satisfies({amiSet: amiSetId})) {
+  if (!req.satisfies({amiSet: AmiSetId})) {
     return;
   }
 
   // Create amiSet
-  let aSet;
+  let amiSetId;
```
jhford: this *will* refer to an AmiSet instance, so let's call it amiSet.

Basically, if a thing refers to a class (or in JS terms a constructor), it should be Caps

Here are some examples:
```javascript
class ExampleClass { }

function ExampleConstructor () { }

let ExampleClass = require('./lib/example-class').ExampleClass;
let exampleInstance = new ExampleClass();

let a = 'john';
```

src/main.js
```diff
@@ -61,27 +60,15 @@ let load = base.loader({
...
 
-  WorkerState: {
+  AmiSet: {
     requires: ['cfg'],
     setup: async ({cfg}) => {
-      let WorkerState = workerState.setup({
+      let AmiSet = AmiSet.setup({
         account: cfg.azure.account,
-        table: cfg.app.workerStateTableName,
+        table: cfg.app.amiSetTableName,
```
jhford: We'll need to add this variable to the config file `config.yml`

src/ami-set.js
```diff
@@ -8,16 +8,37 @@ const KEY_CONST = 'ami-set';
...
-  rowKey: base.Entity.keys.StringKey('amiSetId'),
+  rowKey: base.Entity.keys.StringKey('id'),
 
   properties: {
 
-    amiSetId: base.Entity.types.String,
+    id: base.Entity.types.String,
+    /* This is a JSON object which contains the AMIs of an AMI set keyed by
+     * their virtualization type and region. It is in the shape:
+     * {
```
jhford: This is not specific to your patch, but I'm sort of wondering if a structure like
```json
{
   "us-west-1": {
      "hvm": "ami-111",
      "pv": "ami-222"
  },
   "us-east-1": {
      "hvm": "ami-111",
      "pv": "ami-222"
  }
}
```
or
```json
[
   {
      "region": "us-west-1",
      "hvm": "ami-111",
      "pv": "ami-222"
   },
   {
      "region": "us-east-1",
      "hvm": "ami-111",
      "pv": "ami-222"
   }
]
```
The reason for flipping the hvm/pv around is that the structure goes from most general (region) to more specific (virtualization style) to most specific (ami id).

Whether we have a list of objects or a straigh mapping is something that you'll have to pick.  A list of objects is really nice because you can do really neat things with `Array.prototype.map` and `Array.prototype.filter`.  A mapping is really nice because it's easier to address the values and it is impossible to have duplicates.  Error checking with the object will be easier, but the list of objects is more in line with what we do elsewhere in the provisioner and probably what I'd prefer to see.

src/main.js
```diff
@@ -124,13 +137,15 @@ let load = base.loader({
   },
 
   api: {
-    requires: ['cfg', 'WorkerType', 'Secret', 'ec2', 'stateContainer', 'validator', 'publisher', 'influx'],
-    setup: async ({cfg, WorkerType, Secret, ec2, stateContainer, validator, publisher, influx}) => {
+    requires: ['cfg', 'WorkerType', 'AmiSet', 'Secret', 'ec2', 'validator', 'publisher', 'influx'],
```
jhford: There's another merge issue here.  stateContainer needs to be in both the requires list and the unpacked variable assignments in the setup function signature, but it's not.

src/api-v1.js
```diff
@@ -580,6 +580,94 @@ api.declare({
 
 api.declare({
   method: 'put',
+  route: '/ami-set/:amiSetId',
```
jhford: let's go with just `:id` here as well.  When inside the managing functions for a thing itself, i find it easier to just refer to it as 'id' or something, and use more descriptive names for the ids of things that are referenced as a part of a thing.

src/api-v1.js
```diff
@@ -580,6 +580,94 @@ api.declare({
...
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let AmiSetId = req.params.id;
+
+  input.lastModified = new Date();
+
+  // Authenticate request with parameterized scope
+  if (!req.satisfies({amiSet: AmiSetId})) {
```
jhford: the property that is checked against (here `amiSet`) needs to match the parameterized name in the scopes list (here `amiSetId`, but should be `id`).

After renaming the `amiSetId` variables and scopes to `id`, the object to use here would be `{id: id}`.  In es6+, if your property name and value are derived from the save variable, e.g. `let x = 1; {x: x}` then you can just do `let x = 1; {x}.

src/api-v1.js
```diff
@@ -580,6 +580,94 @@ api.declare({
...
+  stability:  base.API.stability.stable,
+  description: [
+    'Delete an AMI Set.',
+  ].join('\n'),
+}, async function (req, res) {
+  let that = this;
+  let amiSet = req.params.amiSetId;
+
+  try {
+    await this.amiSet.remove({amiSet: amiSetId}, true);
```
jhford: Remember that you're passing in the AmiSet entity class as a context variable to this function in `src/main.js`.  This name here needs to be the name of the context object key, since what we do is take the 'context' object and assign all those values as properties on the `this` in this scope.  In other words, you're looking for `this.AmiSet.remove`.

src/main.js
```diff
@@ -124,13 +137,15 @@ let load = base.loader({
...
+    setup: async ({cfg, WorkerType, AmiSet, Secret, ec2, validator, publisher, influx}) => {
+
       let reportInstanceStarted = series.instanceStarted.reporter(influx);
 
       let router = await v1.setup({
         context: {
           WorkerType: WorkerType,
+          AmiSet: AmiSet,
```
jhford: Regarding my comment about `this.AmiSet.remove`, this is the place where you're creating the value `this.AmiSet`.  This is magic done by the library, but fyi.

src/ami-set.js
djmitche: I tend to agree regarding the list of objects (the second option).  That will also mean we don't need to change our schema when we add a new region.  I also agree that we don't need to solve it in this patch.

jhford:
I'm not sure what the scope of this PR is, but I noticed that there isn't a view or update endpoint for the `AmiSets`.  If that's outside the scope of the PR as you've discussed with @djmitche , it's fine to hold off on that for the time being, but at some point we will need to add these :)

src/main.js
```diff
@@ -59,6 +60,18 @@ let load = base.loader({
     },
   },
 
+  AmiSet: {
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      let AmiSet = AmiSet.setup({
```
jhford: slight bug here.  You are declaring a new variable, `AmiSet`, but assigning the first value to something that uses the variable declared here.  This alone will cause
```
localhost:~/taskcluster/aws-provisioner $ node
> let x = x;
ReferenceError: x is not defined
    at repl:1:31
    at REPLServer.defaultEval (repl.js:252:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:417:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
    at REPLServer.Interface._line (readline.js:549:8)
    at REPLServer.Interface._ttyWrite (readline.js:826:14)
```
What you're looking for here is
```javascript
      let AmiSet = amiSet.setup({ 
```
because you imported the `AmiSet` 'class' using the name `amiSet` above, where you `require()` the file.

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,58 @@
...
+var makeAmiSet = mock.makeAmiSet;
+
+var main = require('../lib/main');
+var helper = require('./helper');
+
+describe('AMI Set api', () => {
+
+  var id = slugid.nice();
+  var amiSetDefinition = makeAmiSet();
+  var amiSetChanged = _.clone(amiSetDefinition);
```
jhford: Without an update endpoint, this variable is not really useful.  For the worker type tests, we have this changed variable where we basically create a second version of the tested definition, with one value changed.  We then use the update method to verify that this expected value matches the one out of the api/table storage.  For now, we should either delete this variable or we should have the update logic implemented.

djmitche:
Yes, we should have "get" and "list" endpoints too.  I feel silly for not noticing that! :)

djmitche:
From talking to @jhford and @anarute today, summarizing the remaining work:
 * get and list endpoints, plus tests
 * schema validation for the `[{region, hvm, pv}, ..]` format

jhford:
Oh, and also an update endpoint.

src/api-v1.js
```diff
@@ -632,6 +632,67 @@ api.declare({
...
+  } catch (err) {
+    if (err.code === 'ResourceNotFound') {
+      res.status(404).json({
+        error: err.code,
+        msg: id + ' not found',
+      });
+    } else {
+      throw err;
+    }
+  }
```
djmitche: This never returns `amiSet`..

src/api-v1.js
```diff
@@ -632,6 +632,67 @@ api.declare({
...
+    let list = await this.AmiSet.listAmiSets();
+    return res.reply(list);
+  } catch (err) {
+    debug('error listing amiSets');
+    debug(err);
+    if (err.stack) {
+      debug(err.stack);
+    }
+    throw err;
+  }
```
djmitche: It looks like elements of `list` here would have a different format than that returned from `getAmiSet`, since `listAmiSets` is using `item.ami.json()`.  The least surprising thing for a user of an API is if the two data structures have the same format.  Using JSON schemas will help enforce this.

src/ami-set.js
```diff
@@ -41,4 +41,29 @@ let AmiSet = base.Entity.configure({
   },
 });
 
+/**
+ * Load all the knwon Amis
+ */
+AmiSet.listAmiSets = async function () {
+
+  let amislist = [];
```
jhford: Normally, we'd camel case this into `amisList`, but given that a list is an aggregation(in this case, a list), I'd normally use a singular for the first part, so `amiList`, but since we're talking about ami sets and not amis, I'd call it `amiSetList`.

src/api-v1.js
jhford: I would almost say that we should just return a list of names, since that's what we do for the worker types.

djmitche: It's done either way in various places in TaskCluster.  I think the key factors are how much work it is to fetch all of the data, and how long the response might get.  In this case, scanning the table to get the names is no less work than scanning the table to get all of the AMI set content; and the resulting response is likely to be a fairly small fixed length (dozens of ami sets), so it's best to return the whole AMI set.

anarute: Should I return `amiSet.amis.json()` in the getAmiSet then?

djmitche: That would refer to a `json()` method on the `amis` property of the azure entity.  Looking in `ami-set.js`, no such method is defined.

It probably makes sense to define an `AmiSet.json()` method which will return the JSON-able value for the whole AMI set -- including `lastModified` and `id`.  Then you can call that method in both getAmiSet and listAmiSets.

schemas/create-ami-set-request.yml
```diff
@@ -0,0 +1,39 @@
...
+        description: |
+          The Amazon AWS region of the AMI set
+      hvm:
+        type: string
+        description: |
+          The AMI that uses HVM virtualization type
+      pv:
+        type: string
+        description: |
+          The AMI that uses PV virtualization type
```
djmitche: you'll want

```
additionalProperties: false
required:
  - region
  - hvm
  - pv
```
here, too.

schemas/get-ami-set-response.yml
```diff
@@ -0,0 +1,39 @@
...
+        description: |
+          The Amazon AWS region of the AMI set
+      hvm:
+        type: string
+        description: |
+          The AMI that uses HVM virtualization type
+      pv:
+        type: string
+        description: |
+          The AMI that uses PV virtualization type
```
djmitche: Same (`additionalProperties`/`required`) here, too.

src/api-v1.js
```diff
@@ -580,6 +580,213 @@ api.declare({
...
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let id = req.params.id;
+
+  input.lastModified = new Date();
+
+  // Authenticate request with parameterized scope
+  if (!req.satisfies({id: id})) {
```
djmitche: (nit) `{id}`

src/api-v1.js
```diff
@@ -580,6 +580,213 @@ api.declare({
...
+  }
+
+  let amiSet = await this.AmiSet.load({id: id});
+
+  await amiSet.modify(function(amiS) {
+    // We know that data that gets to here is valid per-schema
+    for (let key of Object.keys(input)) {
+      amiS[key] = input[key];
+      amiS.lastModified = modDate;
+    }
```
djmitche: There's really only one key of interest here (`amis`).  The `lastModified` doesn't need to be copied from input (and `input.lastModified` doesn't need to be set, above, either).  So while this works, it's a bit over-complicated.

src/api-v1.js
```diff
@@ -580,6 +580,213 @@ api.declare({
...
+});
+
+api.declare({
+  method: 'get',
+  route: '/ami-set/:id',
+  name: 'getAmiSet',
+  output: 'get-ami-set-response.json#',
+  deferAuth: true,
+  scopes: [
+    ['aws-provisioner:view-ami-set:<amiSetId>'],
```
djmitche: Here, `getAmiSet` requires `aws-provisioner:view-ami-set:<id>`, but `listAmiSets` requires no scopes.  It's tricky to enforce scopes in a list method, and I don't see a great reason to hide AMI sets from public view, so I think we should just allow anyone to call `getAmiSet`, too (meaning, remove the `scopes:` here and the `req.satisifies` below).

src/api-v1.js
djmitche: This method is looking great now.

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,77 @@
...
+var makeAmiSet = mock.makeAmiSet;
+
+var main = require('../lib/main');
+var helper = require('./helper');
+
+describe('AMI Set api', () => {
+
+  // var id = slugid.nice();
+  var amiSetDefinition = makeAmiSet();
+  var amiSetChanged = _.clone(amiSetDefinition);
```
djmitche: It doesn't look like this AMI set is actually different from the original.  Maybe add a new AMI to it, just to be more realistic?

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,77 @@
...
+    await client.removeAmiSet(id);
+
+    debug('### Try to load AMI Set');
+    try {
+      await client.getAmiSet(id);
+      throw new Error('Expected and error');
+    } catch (err) {
+      assume(err.statusCode).equals(404);
+    }
+  });
```
djmitche: I don't see a test for `listAmiSets`..

djmitche:
John's review is the one that counts, but as far as I can see this is looking good with just a few more minor things to fix:

 * tests for `listAmiSets`
 * actually change the AMI set in tests
 * schema fixes
 * remove scopes for `getAmiSets`

Once you get going writing tools and backend support for this, you'll probably find some other things to change, but that's easy to do with another PR.

anarute:
Cool! I'm working on this changes and I'll rebase these commits to squash the "fix" ones when I finish them

src/api-v1.js
anarute: I didn't understand, is `{id}` a short form for `{id: id}`?

src/api-v1.js
anarute: should I just remove this logic for now?

src/api-v1.js
djmitche: Yep: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer
```
With ECMAScript 2015, there is a shorter notation available to achieve the same:

var a = "foo", 
    b = 42, 
    c = {};

// Shorthand property names (ES6)
var o = { a, b, c };
```

src/api-v1.js
djmitche: Simplify, yes (basically just `amiS.amis = input.amis`)

test-src/manageamiset_test.js
```diff
@@ -61,6 +67,10 @@ describe('AMI Set api', () => {
     amiSet = await client.getAmiSet(id);
   });
 
+  it('should return a list of AMI sets', async () => {
+    await client.listAmiSets();
+  });
+
```
djmitche: does it actually return a list of AMIs? :)

djmitche: Per out discussion, this should use `assume(amiSets).to.deeply.equal(..)`.

src/ami-set.js
```diff
@@ -0,0 +1,73 @@
...
+ * Load all the knwon Amis
+ */
+AmiSet.listAmiSets = async function () {
+
+  let amiSetList = [];
+
+  try {
+    await base.Entity.scan.call(this, {}, {
+      handler: function(item) {
+        amiSetList.push(item.json());
```
jhford: If this is a listing method, we should be pushing the item.id property instead of the complete item in a json-serializable format.  If the goal is a load-all method, let's call it `loadAll()` to match the worker type [here](https://github.com/taskcluster/aws-provisioner/blob/master/src/worker-type.js#L249)

src/api-v1.js
```diff
@@ -580,6 +580,197 @@ api.declare({
...
+      throw err;
+    }
+    amiSet = await this.AmiSet.load({id});
+
+    // Check if it matches the existing amiSet
+    let match = [
+      'id',
+      'amis',
+    ].every((key) => {
+      return _.isEqual(amiSet[key], input[key]);
```
jhford: I can't remember off hand, but we should ensure that `_.isEqual` is a deep equality operator, since an amiSet is an object which collects other objects.

src/api-v1.js
```diff
@@ -580,6 +580,197 @@ api.declare({
...
+  }
+  res.reply({outcome: 'success'});
+  return;
+
+});
+
+api.declare({
+  method: 'get',
+  route: '/ami-set/:id',
+  name: 'getAmiSet',
```
jhford: The `WorkerType` entity uses `name:  'workerType'`.  Can we follow that pattern here?

src/api-v1.js
```diff
@@ -580,6 +580,197 @@ api.declare({
...
+  let input = req.body;
+  let id = req.params.id;
+
+  if (!req.satisfies({id})) {
+    return;
+  }
+
+  let amiSet = await this.AmiSet.load({id: id});
+
+  await amiSet.modify(function(amiS) {
```
jhford: the name `amiS` sort of implies that it's a plural of amis but with confusing captalization.  Instead, since we're passing in an instance of an AmiSet, let's use the `amiSet` variable name.

src/api-v1.js
jhford: The provisioner mostly uses list of strings for list methods.  I'd like to stick to that here unless there's a compelling reason not to.  That would also make having a json-schema for this endpoint pretty quick to write up.

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,144 @@
...
+  let client;
+
+  before(async () => {
+    amiSet = await main('AmiSet', {process: 'AmiSet', profile: 'test'});
+
+    client = helper.getClient();
+  });
+
+  beforeEach(async () => {
+    await main('tableCleaner', {process: 'tableCleaner', profile: 'test'});
```
jhford: We should add the `AmiSet` entity to the `tableCleaner` component in `src/main.js` so that this table is actually cleaned out.

src/ami-set.js
djmitche: I believe we already had some discussion of this in the PR?  Many of the list methods in TC services return the entire object, e.g., [auth.listClients](https://docs.taskcluster.net/reference/platform/auth/api-docs#listClients).  The reason *not* to do so is if it requires additional queries or a lot more bandwidth.  Neither is the case here.  We have the additional information about each AMI set returned from Azure, so why not provide it to the user?

src/api-v1.js
djmitche: [It is.](https://lodash.com/docs#isEqual)

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,144 @@
...
+    } catch (e) {
+      console.log(JSON.stringify(e));
+      throw e;
+    }
+
+    debug('### Load amiSet (again)');
+    amiSet = await client.getAmiSet(id);
+    lastModified = amiSet.lastModified;
+    assume(await amiSet).to.deeply.equal({
+      amis: [
```
jhford: we use mock objects in a single file elsewhere to make it easier to update the format of the test objects without having to completely rewrite the test.  In this case, the objects are simple enough that I don't know if it's worth doing here, but please take a look at the `test-src/mock*.js` files just to see how they're structured.  Basically, we have a default that is returned, but you can overwrite the values you want overwritten.  Not critical here, but worth looking at.

anarute:
@jhford I fixed the things you've mentioned:
- added AmiSet table to tablecleaner;
- fixed naming patterns;
- cleaned the amiSet tests using only data from mockamiset_test.js - except data for comparison, like here https://github.com/taskcluster/aws-provisioner/pull/85/files#diff-bf33c71f7b5151be915935e33c040095R66
- changed `listAmiSets` method to return the names of all amiSets, not the objects. Should I create a `loadAll()` method for amiSets?

djmitche:
If we need `loadAll()`, it's easy enough to add in a new PR :)

jhford:
So there's a problem here.  The create ami set endpoint has a json-schema that forces there to be an ID parameter in the request body.  The problem is that we already have that specified in the parameters of the api method.  The create schema should only have the amis object.

schemas/get-ami-set-response.yml
jhford: wait... why?  pv is a type of string, it doesn't have properties... am I missing something?

jhford:
Ok.  So I think the only issue left here is removing the createAmiSet json-schema allowing an ID value.  The only place we should be supplying the id value is the parameter that we bake into the `/ami-set/:id` route.

I think with that change, i'm 99% sure that we're ready to merge!

schemas/get-ami-set-response.yml
djmitche: Oh, indentation is a bit funny here, too -- it says `amis` has properties, rather than the items in `amis`.  My comment was indicating that the region, hvm, and pv properties should be limited and all required.

jhford: ahh right, so just intending it a bit.  `amis` is an array, so it shouldn't have any properties.  I'm surprised this schema works

jhford: http://jsonschemalint.com/draft4/ is your friend.  I patched it to support YAML serialised schemas and documents

anarute:
@jhford I removed the id property from `createAmiSet` schema, could you review the code again, please?

schemas/create-ami-set-request.yml
```diff
@@ -0,0 +1,39 @@
...
+        pv:
+          type: string
+          description: |
+            The AMI that uses PV virtualization type
+      additionalProperties: false
+      required:
+        - region
+        - hvm
+        - pv
+  lastModified:
```
jhford: Sorry that I didn't see this before, but we also should not have last modified as an allowed field in the request.  we will generate this value inside the api handler, so forcing a value here that we won't end up using isn't ideal.

src/api-v1.js
```diff
@@ -580,6 +580,197 @@ api.declare({
...
+  title: 'Create new AMI Set',
+  stability:  base.API.stability.stable,
+  description: [
+    'Create an AMI Set. An AMI Set is a collection of AMIs with a single name.',
+  ].join('\n'),
+}, async function (req, res) {
+  let input = req.body;
+  let id = req.params.id;
+
+  input.lastModified = new Date();
```
jhford: We shouldn't really be setting properties on the request body object.  Also, this value isn't actually used in the `AmiSet.create` method call below

src/api-v1.js
```diff
@@ -580,6 +580,197 @@ api.declare({
...
+    return;
+  }
+
+  // Create amiSet
+  let amiSet;
+  try {
+    amiSet = await this.AmiSet.create({
+      id: id,
+      amis: input.amis,
+      lastModified: new Date(),
```
jhford: here's where we're actually using this value

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,107 @@
+var slugid = require('slugid');
```
jhford: random nit, this file uses `let` and `var`.  We should probably chose one and stick to it.

The difference in case you're unfamiliar is that var is in scope through the entire function.  If you use var in an `if`, the variable is in scope through the entire function.  Let is block scope, if you use let in an `if`, it's only in scope inside that `if` and those blocks (e.g. code inside of `{ }`) that are nested.

Let is better than var!  Also, if you're really curious do a search about 'variable hoisting' for more info on how var works.

test-src/manageamiset_test.js
```diff
@@ -0,0 +1,107 @@
...
+      await client.updateAmiSet(id, amiSetChanged);
+    } catch (e) {
+      console.log(JSON.stringify(e));
+      throw e;
+    }
+
+    debug('### Load amiSet (again)');
+    amiSet = await client.amiSet(id);
+    lastModified = amiSet.lastModified;
+    assume(await amiSet).to.deeply.equal({
```
jhford: we don't need to await `amiSet` because we already await'd it above.

jhford: Also, i think this is way outside the scope of this project, but one thing that I like to do is compare the amiSetChanged object itself.  Doing that does require that you handle the lastModified stuff, as you do correctly below.  Another option is to use Sinon to mock out the system clock.  That's really out of scope and what you have here is fine.  I might consider using `_.default` to grab the amis object from amiSetChanged

src/api-v1.js
anarute: you're right, I passed that out among all the changes! thanks for noticing

test-src/manageamiset_test.js
anarute: thanks for the explanation!

PR #86, opened 2016-06-08
improvements

jhford:
working on some general improvements here

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -417,7 +417,6 @@ api.declare({
       missing: missing,
     });
   }
-
   input.lastModified = modDate;
 
   if (!req.satisfies({workerType: workerType})) { return undefined; }
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -7,6 +7,7 @@ let series = require('./influx-series');
 let keyPairs = require('./key-pairs');
 let _ = require('lodash');
 let delayer = require('./delayer');
+let amiExists = require('./check-for-ami');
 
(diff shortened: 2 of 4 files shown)


jhford:
`dont do a watchdog kill while we're actively submitting requests` is to take care of an edge case where we are bumping maxCapacity of a very behind worker type to work the backlog.  The idea is to make sure that the watchdog does it's thing properly, but now we're going to consider submitting a spot request successfully to be a good enough event to touch the watch dog.

jhford:
`bug 1264956 - don't try to provision worker types that have an invalid ami` is to make sure that having a missing AMI doesn't cause the provisioner to fail iterations.  I'd like to set something up where I automatically inform the owner of the worker type that their worker type is invalid.  Unlike errors that happen while creating or updating worker types, this error can start on a non-user-triggered action.  That means that we need out of band notifications.  Emails to owners is outside the scope of this.  See the commit header for an example of what happens when an invalid AMI is specified.

jonasfj:
r+

I see you left todos to submit to sentry...

jhford:
15f3d0e was merged but  5bcd621 had issues and was backed out.  Closing this PR for now and will retry the second commit another time.

selenamarie:
Reopening for further hacking! @djmitche

jhford:
This is being addressed with https://github.com/taskcluster/aws-provisioner/pull/95.  Not sure if we need both of these prs

djmitche:
agreed

PR #87, opened 2016-06-21
switch to structured logging library

jhford:
This uses the bunyan-based taskcluster-lib-log library with its debugCompat function to bridge the whole aws-provisioner from the debug module over to the bunyan based logging system.  This is not an attempt to break each message down into the correct logging levels, rather that can be done at a later time now that we have the library hooked up and ready to go.  All of the modules import a single simple ./src/log.js module to be able to share the root logger.  Sharing the root logger isn't mandatory, but it is nice from a hierarchy standpoint as it means that all messages will decend from the same config.  I did this instead of putting it into src/main.js because i'd like to have each module have a module-level psuedo-root-logger available to make logging less of a pain.


jhford:
Note that the intention here is that to get nicer looking log files, we'll end up doing something like "heroku logs -t | bunyan -o short" or something similar... we could do the bunyan post-processing in the procfile, but then we lose all the structure from the logs.

djmitche:
I don't see `src/log.js` here -- does it need to be git add'd?

jhford:
I always do that! sorry.

djmitche:
:+1:

PR #88, opened 2016-06-21
User runnable

djmitche:
@anarute

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ To run the provisioner locally, you will need a similar set of configuration.
 Then run
 
 ```
-npm run compile && DEBUG=* node lib/main.js development
+npm run compile && DEBUG=* node lib/main.js server
 ```
 
 Note that this is not usually the best way to test the provisioner.  Be careful
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ To run the provisioner locally, you will need a similar set of configuration.
 Then run
 
 ```
-npm run compile && DEBUG=* node lib/main.js server
+npm run compile && NODE_ENV=development DEBUG=* node lib/main.js server # just web server
+npm run compile && NODE_ENV=development DEBUG=* node lib/main.js all    # web + provisioner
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -46,7 +29,7 @@ To run the provisioner locally, you will need a similar set of configuration.
 Then run
 
 ```
-npm run compile && DEBUG=* node lib/main.js development
+npm run compile && DEBUG=* node lib/main.js server
```
jhford: s/server/all/ and you can have the complete provisioner running and not just the server.  We should also do `NODE_ENV=development`

config.yml
```diff
@@ -83,6 +83,7 @@ development:
   azure:
     account: jungle
   app:
+    awsInstancePubkey: 'fake-pubkey'
```
jhford: i don't know if it's OK to use a fake key in this format, a valid-shaped key should be in user-config.yml instead.

src/main.js
```diff
@@ -201,11 +201,15 @@ let load = base.loader({
...
-      });
+      if (cfg.influx.connectionString) {
+        return new base.stats.Influx({
+          connectionString: cfg.influx.connectionString,
+          maxDelay: cfg.influx.maxDelay,
+          maxPendingPoints: cfg.influx.maxPendingPoints,
+        });
+      } else {
+        console.log("No influx.connectionString configured; not using influx");
+      }
```
jhford: Shouldn't we use NullDrain instead?  I'm not sure if influx being undefined will be OK for the provisioning background process.

config.yml
djmitche: The field is required, but if you're not running the provisioner, it's not actually used.  So this gets the web server to run without having to create a key.

djmitche:
Does this look better?

jhford:
This looks fine.  I'm not merging today because it's 6:30pm and I don't want to watch the deployment today, but I will deploy it tomorrow.

PR #91, opened 2016-07-20
move to taskcluster-lib-iterate

jhford:
Here's an example implementation of lib-iterate in the provisioner.  It's definitely looping here, but I'll let it go over night to double check

@imbstack


src/main.js
```diff
@@ -260,19 +261,42 @@ let load = base.loader({
...
+          apiKey: cfg.deadmanssnitch.api.key,
+          snitchUrl: cfg.deadmanssnitch.iterationSnitch,
+        },
+        handler: async (watchdog, state) => {
+          await provisioner.runAllProvisionersOnce();
+        },
+      });
+      i.start();
+      return new Promise((res, rej) => {
+        i.on('started', () => {
```
jhford: i almost wonder if this shouldn't be stopped... that way you'd basically have a promise that lasts as long as the provisioner does, which might be nice.

src/main.js
```diff
@@ -260,19 +261,42 @@ let load = base.loader({
...
+      let i = new Iterate({
+        maxIterationTime: 1000 * 60 * 10,
+        watchDog: 1000 * 60 * 10,
+        maxFailures: 15,
+        waitTime: cfg.app.iterationInterval,
+        dmsConfig: {
+          apiKey: cfg.deadmanssnitch.api.key,
+          snitchUrl: cfg.deadmanssnitch.iterationSnitch,
+        },
+        handler: async (watchdog, state) => {
```
jonasfj: what is state and watchdog?

jhford: watchdog is the same thing we used to have in the provisioner directly.  think of it as a ticking timebomb.  We light the fuse each time we start an iteration.  the `.touch()` method resets the fuse.  If we let the fuse burn out before either calling `.touch()` or `.stop()` on it, it causes an error.

There are two watchdogs in lib-iterate.  one is an overall iteration watchdog, that one basically says an iteration can never ever take more than this amount of time.  A second one is passed into the handler function that can be `.touch()`'d each time work happens.  This lets us say "an iteration can take no longer than 20 minutes under any circumstance, but if we aren't doing work every 10s, we know we've frozen already".

An example is in the provisioner when we're submitting bids.  We work through that list, each submission takes .5s or 1.5s, and in total can take a while, so we know that if we haven't made a submission in >1m, we're frozen.

jhford:
I ran this overnight on a development instance and it worked.

jhford:
merged in a different version of this patch.

PR #92, opened 2016-07-21
Bug 1244656 - Long live terminateAllInstancesOfWorkerType and shutdownEverySingleEc2InstanceManagedByThisProvisioner

petemoore:
That should do it!

Brought to life from the [graveyard](https://github.com/taskcluster/aws-provisioner/blob/cf460f8846df79833b2accd44b15f89a21c01e8c/lib/routes/v1.js#L575-L663) but with slightly adjusted scopes.

Untested.

CC @djmitche @jonasfj

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -1099,3 +1099,93 @@ api.declare({
     lastCheckedIn: snitch.checked_in_at,
   });
 });
+
+api.declare({
+  method: 'post',
+  route: '/worker-type/:workerType/terminate-all-instances',
+  name: 'terminateAllInstancesOfWorkerType',
+  title: 'Shutdown Every Ec2 Instance of this Worker Type',
+  stability:  base.API.stability.experimental,
+  scopes: [
+    [
+      'aws-provisioner:terminate-all-worker-type:<workerType>',
+    ],
+  ],
+  description: [
+    'WARNING: YOU ALMOST CERTAINLY DO NOT WANT TO USE THIS ',
+    'Shut down every single EC2 instance associated with this workerType. ',
+    'This means every single last one.  You probably don\'t want to use ',
+    'this method, which is why it has an obnoxious name.  Don\'t even try ',
+    'to claim you didn\'t know what this method does!',
+    '',
+    '**This API end-point is experimental and may be subject to change without warning.**',
+  ].join('\n'),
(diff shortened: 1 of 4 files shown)


src/api-v1.js
```diff
@@ -1099,3 +1099,93 @@ api.declare({
     lastCheckedIn: snitch.checked_in_at,
   });
 });
+
+api.declare({
+  method: 'post',
+  route: '/worker-type/:workerType/terminate-all-instances',
+  name: 'terminateAllInstancesOfWorkerType',
```
djmitche: how about `terminateWorkerTypeInstances`

src/api-v1.js
```diff
@@ -1099,3 +1099,93 @@ api.declare({
...
+  });
+
+  return p;
+
+});
+
+api.declare({
+  method: 'post',
+  route: '/shutdown/every/single/ec2/instance/managed/by/this/provisioner',
+  name: 'shutdownEverySingleEc2InstanceManagedByThisProvisioner',
```
djmitche: `terminateAllInstances`

djmitche: And I would argue, we don't need this endpoint back - can you think of a time we have needed it?

jhford: nope.  these names are intentionally long, annoying to type and very descriptive.  There should be as many speed bumps as possible to their use.

djmitche: I strongly disagree with that as a design tactic.

Twisted did something similar with `twisted.cred.InMemoryUsernamePasswordDatabaseDontUse` which is used in just about every use of cred that I've ever seen (since they are not, as the designer imagined, supporting full user applications, but rather simple HTTP APIs using some kind of machine-based authentication).

If someone needs to use this endpoint, they are not going to say "boy that seemed appealing but it's so much to type so I guess I'll just login to the console instead".  And if they don't need it, they won't use it.  The documentation is sufficiently dissuasive.

jonasfj: @jhford the description property of the documentation supports markdown... So you can insert a lot of **bold** text... Underlining... or a dramatically ugly image showing a guns shooting off a someones foot (maybe that's too graphic).

My point is formatted text in docs might be a good alternative... IMO we don't need the end-point though...

PR #94, opened 2016-07-27
Only submit the first 400 requests

jhford:
This will mean that only the first 400 bids are made.  We shuffle the list of submissions to make in a statistically random way.  Because the provsioner is not able to make any reasonable judgement about which worker types are more important, nor whether any given capacity unit is more important than any other, this is the best approach.

Limiting us to 400 instances being requested per iterations does a couple different things for us.  First, we will get status updates to the provisioner ui more quickly.  We'll also get fewer deadmans snitch warnings.

More importantly, however, is that as we work through the backlog, we can reevaluate which things should really start.  Since an unbounded list could take over an hour, we could have worked through the backlog by the time the provisioner iteration completes.

### DIFF
--- a/src/provision.js
+++ b/src/provision.js
@@ -296,6 +296,11 @@ class Provisioner {
     // any particular worker type
     forSpawning = shuffle.knuthShuffle(forSpawning);
 
+    // We'll only consider the first 400 requests.  Any that are dropped on the
+    // floor will be computed on the next iteration and have an equal chance of
+    // being submitted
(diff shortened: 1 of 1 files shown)


jonasfj:
r+

jonasfj:
This seems like a potential huge improvement... But if this is 1000 requests from a single workerType maybe random shuffle isn't good enough  anymore... ie. maybe it has to be round-robin...

---

huge thanks for picking this up!!!

PR #95, opened 2016-07-28
bug 1264956 check that each worker type that we're going to try to spawn is valid

jhford:
While we already have guards in the API against this, it does not protect against modification of the critical data fields after the fact.  There is still a possibility of something going wrong between when an iteration starts and when the submission is made, but this is a very small window and will only last for at most one iteration after the bad state occurs.

Future idea is that we update the WorkerType definition or State (or create something new) to store the list of worker types that are invalid and the reason for their invalidity.

### DIFF
--- a/src/provision.js
+++ b/src/provision.js
@@ -296,11 +296,6 @@ class Provisioner {
     // any particular worker type
     forSpawning = shuffle.knuthShuffle(forSpawning);
 
-    // We'll only consider the first 400 requests.  Any that are dropped on the
-    // floor will be computed on the next iteration and have an equal chance of
-    // being submitted
-    forSpawning = forSpawning.slice(0, 400);
-
     // Let's find the unique worker types in the list of requests to make
     let toSpawnWorkerTypes = _.intersection(forSpawning.map(x => x.workerType.workerType));
 
@@ -315,6 +310,11 @@ class Provisioner {
     }
 
     forSpawning = forSpawning.filter(x => disabled.indexOf(x.workerType) === -1);
+
+    // We'll only consider the first 400 requests.  Any that are dropped on the
+    // floor will be computed on the next iteration and have an equal chance of
+    // being submitted
+    forSpawning = forSpawning.slice(0, 400);
 
     while (forSpawning.length > 0 && attemptsLeft-- > 0) {
       let toSpawn = forSpawning.shift();
--- a/src/provision.js
+++ b/src/provision.js
@@ -111,7 +111,7 @@ class Provisioner {
     try {
       do {
         debug('starting iteration %d, consecutive failures %d',
-              this.__stats.runs, this.__stats.consecFail);
+            this.__stats.runs, this.__stats.consecFail);
 
         // If we don't do this, we'll have an uncaught exception
         this.__watchDog.touch();
@@ -275,7 +275,7 @@ class Provisioner {
         debug('destroying %d for %s', capToKill, worker.workerType);
         try {
           await this.awsManager.killCapacityOfWorkerType(
-                worker, capToKill, ['pending', 'spotReq']);
+              worker, capToKill, ['pending', 'spotReq']);
         } catch (killCapErr) {
           debug('[alert-operator] error running the capacity killer');
           debug(killCapErr.stack || killCapErr);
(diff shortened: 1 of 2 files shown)


jonasfj:
@jhford, huge apologies for some reason this got hidden in a bunch of emails...

I'll review right away :)

src/provision.js
```diff
@@ -296,6 +296,21 @@ class Provisioner {
     // any particular worker type
     forSpawning = shuffle.knuthShuffle(forSpawning);
 
+    // Let's find the unique worker types in the list of requests to make
+    let toSpawnWorkerTypes = _.intersection(forSpawning.map(x => x.workerType.workerType));
```
jonasfj: Why `_.intersection` are you looking for `_.uniq`...?

src/provision.js
```diff
@@ -296,6 +296,21 @@ class Provisioner {
...
+
+    let disabled = [];
+    for (let toTest of workerTypes.filter(x => toSpawnWorkerTypes.indexOf(x) !== -1)) {
+      // Consider iterating over a list where we filter out those worker types
+      // which don't have outstanding requests
+      let canLaunch = await this.awsManager.workerTypeCanLaunch(toTest);
+      if (!canLaunch) {
+        disabled.push(toTest.workerType);
+      }
```
jonasfj: May I suggest:
```js
if (!await this.awsManager.workerTypeCanLaunch(toTest)) {
  forSpawning = forSpawning.filter(x => x.workerType !== toTest.workerType);
}
```

jonasfj: Getting rid of `disabled` would certainly help... 

I wish I saw a more elegant way to do this... but it seems okay to me.

jonasfj:
# `r+`

I suggest you consider comments...

----
So thinking a bit about this... I'm wondering if maybe this isn't the right approach..

It's a bit like when people try to tell us to not do:
```
if (file_exists(path)) {
  text = read_file(path)
} else {
  text = "fallback-value";
}
```
but instead tell us to do:
```
try{
  text = read_file(path)
} catch (err) {
  if (err.code != "FileNotFound") {
    throw err;
  }
  text = "fallback-value";
}
```
....


-----

So what I'm saying is that instead of trying to predict an error, we should attempt to do what we want to do... and handle the error instead..

In the file example we have race condition where someone could delete the file between the `file_exists(path)` call and the `read_file(path)` call... In this code we have race condition where someone could delete the AMI while provisioning is going on...

Perhaps it's better to just try to provision and when an error, we analyse the error.. If the error is due to the launchSpecification (ie. it's not a 5xx error, but a 4xx error), then we filter out the rest of the requests for this `workerType`... So that we won't attempt to provision those...

Sorry, for the long note... tl;dr: Let's handle 4xx errors and filter after an error happened rather than trying to avoid errors.

jonasfj:
Perhaps: https://github.com/taskcluster/aws-provisioner/pull/99

This is a more robust way to approach this... I haven't actually looked the errors... But handling them in a smart way might be safer... We can't stop garndt and dustin from deleting AMI at the most inconvenient times :)

jhford:
So i'm not doing 
```
if (file_exists(path)) {
  text = read_file(path)
} else {
  text = "fallback-value";
}
```
but something closer to
```
if (hard_drive_exists(path)) {
  for (let x of listDir(path) {
    try {
      text = read_file(path)
    } catch (err) {
      if (err.code != "FileNotFound") {
        throw err;
      }
      text = "fallback-value";
     }
  }
}
```

I'm verifying that each worker type that we're even considering spawning is a valid definition and using that outcome to decide whether *any* of these worker types should be attempted.

jonasfj:
ah, that's partly true... except you actually dont filter out workerTypes
that are broken, if creation causes an error.... look at the PR i made to
illustrate it I suspect that might be a simple fix.. ie look at the error
to see if it is permanent or not...

Den 5. aug. 2016 6.23 AM skrev "John Ford" <notifications@github.com>:

jhford:
Yep.  because if I *think* it's valid, then any failure to provision should be considered a real failure, and in this case, any failure reason is a valid reason.  The case that I'm trying to avoid is that I get 100 requests to provision, then 100 fail, and I say "it's all good" and continue.  Instead, this patch is trying to say "I have 100 requests, but 30 are for a worker type I know won't work, so I wont submit those, but if I have less than 70 success from the other, everything worked!"

I've always been conflicted about how to treat failure of the actual requestSpotInstancce method :/  I'm goign to do a lot of thought on that for the DO provisioner

jonasfj:
> Instead, this patch is trying to say "I have 100 requests, but 30 are for a worker type I know won't work, so I wont submit those, but if I have less than 70 success from the other, everything worked!"

Ah, so the argument I'm making is that you try the 100 request and when one fails with a 4xx error you filter out the 29 other requests for the workerType that failed and say they won't work...  I assume you're trying to avoid making another 29 requests that you expect will return 4xx...

hmm... Maybe the solution is to send some serious alerts if we get 4xx errors. But only alert once per workerType per 24 hours (ie. cache in provisioner memory that we've already alerted on a workerType).
When bstack finishes tc-notify, it might just a tc scope and a quick API call to send an email to `workerType.owner` (when people start setting the property, that'll be useful)...

jhford:
I just realised that I should be using this same method in the API as well for create and update checks.  To do this I'll need to change the can launch check function to use the WorkerType.createLaunchSpec class method instead of the instance version, or at least be able to work on just a flat object version of a worker type.  This is so we can test the input from the user of the API without actually creating the entity.

jhford:
@jonasfj as for the error stuff.... we could just submit all errors to sentry and ignore the failure...  Sentry should have some smarts about only alerting every now and then right?

jonasfj:
@jhford, yes sentry will only alert on new errors and merge matching errors... In fact it rarely alerts mostly it just collects the errors... (or maybe I show how disabled my alerts).

jhford:
This got merged

PR #97, opened 2016-07-29
Fix scopes for getting workerType state

djmitche:
This should clear things up so that users can view the status for garbage workerTypes.


anarute:
Tested it locally and it works for me :)

djmitche:
Thank you!

PR #98, opened 2016-08-01
bug 1290922 - stop using KeyPair.Name as a metadata store

jhford:
For security reasons, we'd like to use a different mechanism to store the provisioner id and worker type.  This approach builds up a database in memory, saving to Azure blob storage as needed.

When this deploys, we'll stop considering worker types which are managed using KeyPairs.  This will result in over provisioning briefly, but the worker shutdown procedures should ensure that this overlap is not too much more than one hour.


jonasfj:
this code has more deletions than additions, I guess we should approve it based on that alone...

src/aws-manager.js
```diff
@@ -267,7 +256,15 @@ class AwsManager {
         // Now let's classify them
         for (let reservation of response[0].data.Reservations) {
           for (let instance of reservation.Instances) {
-            let workerType = this.parseKeyPairName(instance.KeyName).workerType;
+            // We want to be able to transition from old (key name) to the new
+            // system based on spot request ids
+            if (!instance.SpotInstanceRequestId) {
```
jonasfj: change comment to say... ignore ondemand instances...

src/aws-manager.js
```diff
@@ -267,7 +256,15 @@ class AwsManager {
...
+              continue;
+            }
+            let workerType = await this.workerTypeForSRID(instance.SpotInstanceRequestId);
+            if (!workerType) {
+              continue;
```
jonasfj: This is where you would do the userData look...

src/aws-manager.js
```diff
@@ -267,7 +256,15 @@ class AwsManager {
         // Now let's classify them
         for (let reservation of response[0].data.Reservations) {
           for (let instance of reservation.Instances) {
-            let workerType = this.parseKeyPairName(instance.KeyName).workerType;
+            // We want to be able to transition from old (key name) to the new
+            // system based on spot request ids
+            if (!instance.SpotInstanceRequestId) {
+              continue;
+            }
+            let workerType = await this.workerTypeForSRID(instance.SpotInstanceRequestId);
```
jonasfj: change `workerTypeForSRID` to be `determineWorkerType(instance)`... then  `determineWorkerType(instance)` will find the workerType based on `spotRequestId` or by looking up instance id...

And `determineWorkerType(instance)` would return `null`, if it's not managed by the aws-provisioner...

jhford:
@jonasfj hey, mind taking another look at this.  I think this solves the issue pretty well.  We no longer have the bad case of looking up userData for every single instance in the account.  It's also persisting the EC2 responses that take a while to show up in the API's describe* calls.  About time!  This means that a provisioner restart is now literally a non-op, we shouldn't lose any important imformation.  I'm thinking of moving the state persistence into a different class, but for now I think this is good enough.

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
     this.reportAmiUsage = series.amiUsage.reporter(influx);
   }
 
+  async init(){
+    try {
+      let rawJson = await this.spotRequestContainer.read('internal-provisioner-data');
+      let data = JSON.parse(rawJson);
+      // This is mainly for debugging
+      fs.writeFileSync('internal-data.json', data);
```
jonasfj: Should we perhaps avoid doing this in production... Or at the very least use a non-sync writefile..

jonasfj: package `mz/fs` is pretty cool...

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+      // If we've known about this for more than 5 days, evict it, otherwise
+      // keep it.  5 days is a good amount since we force kill things even if
+      // they're running after 4 days.  This lets us have a full extra day
+      if (diff > 1000 * 60 * 60 * 24 * 5) {
+        return false;
+      }
+      return true;
+    }
+
+    this.__spotRequestIdCache = this.__spotRequestIdCache.filter(expire);
```
jonasfj: We only use `expire here... Can't we make it arrow function for readability...
```js
// Expire spot requests ids from state after 5 days, ensuring an extra day after force kill
const stateExpiration = 1000 * 60 * 60 * 24 * 5;

this.__spotRequestIdCache = this.__spotRequestIdCache.filter(item => Date.now() - item.created < stateExpiration);
```

jonasfj: @jhford, are spot request ids never reused by EC2? I know that machine instance ids are reused...

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+      managedSpotRequests: this.__spotRequestIdCache || [],
+      internallyTrackedSpotRequests: this.__internalState || [],
+    }, null, 2);
+
+    fs.writeFileSync('internal-data.json', data);
+    return this.spotRequestContainer.write('internal-provisioner-data', data);
+  }
+
+  /**
+   * Simplify how we access stored spot request ids
```
jonasfj: Maybe a comment saying what the function takes and returns...

jonasfj: nit

jonasfj: "Given resource and region workerTypeForResource returns the workerType that the resource was created for".

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+    assert(typeof resource === 'object');
+    assert(typeof region === 'string');
+
+    let srid = resource.SpotInstanceRequestId;
+    let instanceId = resource.InstanceId;
+
+    // First, we should check to see if we have the mapping of
+    // SpotInstanceRequestId in memory (which we persist to an azure blob
+    // storage blob).  If this item is in that cache, we can be sure that the
+    // provisioner both owns it because it created it, and also which worker
```
jonasfj: we know we own it because it has the provisioners keyName.

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+    // type, since we add that information to the cache when we add the id.
+    let found = this.__spotRequestIdCache.filter(x => x.id === srid && x.region === region);
+    if (found.length === 1) {
+      return found[0].workerType;
+    } else if (found.length > 1) {
+      let firstX = found[0].workerType;
```
jonasfj: Isn't it an invariant violation if there is more than one?

IMO we should just do: 
```js
let entry = this.__spotRequestIdCache.filter(x => x.id === srid && x.region === region).pop();
if (entry) {
  return entry.workerType;
}
```

And when we load `__spotRequestIdCache` as well as after each iteration we do some sanity checks. To ensure the invariants are enforced.
Example:
```js
let uniqueEntries = _.uniqBy(this.__spotRequestIdCache, entry => entry.region + '/' +entry.id).length;
assert(uniqueEntries === this.__spotRequestIdCache.length, 'Duplicate entries in __spotRequestIdCache');
```

This is much less code and enforces our invariants much better... If we include the assert above after loading state, before saving state (so we never save bad state), and after modifying the state in an iteration.. Then we're in a really good place.

We could probably just introduce a simple `this._validateState()` method that validates all the internal state.. Sends the an error to sentry if there is a violation... Then we can also use `this._validateState()` to drop current state if there is an invariant violation. We can do this simple thing after every iteration, after/before loading/saving state, etc... 

Having a few check points where we validate all invariants of the internal state is a lot simpler.

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+      }
+    }
+ 
+    // If we have no instance ID at this point, we cannot do any further
+    // checking to see which WorkerType this belongs to.  As well, since
+    // unfulfilled spot request don't cost us money, we're not worried about
+    // rouges costing money.
+    // TODO: Consider using DescribeTags with filters to see if this instance
+    // or spot request has already been tagged.  Also tag things as part of
+    // requestSpotInstance
```
jonasfj: If the resource is a spotRequest then the describeSpotRequest result contains the tags.. So we should able to just look at those... Assuming that we're in fact tagging spot requests.

src/aws-manager.js
```diff
@@ -139,10 +125,172 @@ class AwsManager {
...
+    if (!instanceId) {
+      return null;
+    }
+
+    // If we're here, we have not found a result.  This means that we should
+    // try to load the UserData if we can.  We do this because there's a chance
```
jonasfj: Don't you need to check the instance tags... They should be in on the `resource` object... assuming it's an object that is returned from `describeInstances`

src/aws-manager.js
jonasfj: Note: You should never have to call `DescribeTags` as it's already in the response from describeInstances/describeSpotRequests.

jonasfj:
See comments. there is some refactoring for invariant validation that easy and worth doing...
Also you're not looking at the tags...

src/aws-manager.js
jhford: true, you're right.

src/aws-manager.js
jhford: I think that's reasonable.  I just don't want to cause the provisioner to barf if an identical item is in there twice.  I'll look at this shortly.

src/aws-manager.js
jhford: yep, we know we own it for both reasons, though.  In this case, we're using the ownership knowledge gleaned from having been the provisioner to submit it, which is why that's what the comment says :)  But yes, we own it for both reasons :)

src/aws-manager.js
jhford: yes.  agreed.

src/aws-manager.js
jhford: So they are, I'm 99% sure.  The last time we talked about this, it was a security issue around worker-ids, if I'm recalling correctly.  In that case, the fear was that an unprivileged worker type could get a worker id for a wrong instance type.  For security issues like that, I think it was the right call to not do something because of that edge case.

This case is a lot different, however.  The worst case is that if we had a spot instance reuse, we'd count an instance for the wrong worker type for provisioning and if the wrong worker type were deleted, we might accidentally kill it.  This isn't a security issue, but sadly, it's basically the only way for us to link a spot request to a worker type in a reliable way :(

We'd also likely only be dealing with one or two instance if this happened, and we evict SRIDs after 5 days, so we'd need to have a signficant amount of reuse within the same region within the same 5 day period, and the worst case is that we are off on provisioning by at most the amount of capacity of the instances with reused spot instance request ids.

I can live with this risk, how about you?

src/aws-manager.js
jhford: hmm.  i guess it really was only there for development.  I could wrap it in an `if (process.env.NODE_ENV !== 'production')`, but i'm leaning towards just putting it in a log.trace.  I don't want to pass the profile into the actual aws-manager.js file because I think that's making it easier to do bad things based on environment, when we should have all that as config stuff....

The more i think about it, log.trace is likely what i'll do.

src/aws-manager.js
jhford: regarding expire, that being a function is leftover from when I was tracking a list of unmanaged instance ids, and the data structures both had a .created property.  I think we're good to reduce complexity here

jhford:
Not looking at tags was semi-intentional.... I think it's probably worthwhile to implement that.  Besides the invariant stuff, do you like this overall?

src/aws-manager.js
jonasfj: Yeah, seems like an acceptable risk to me..

jonasfj:
So I think we should look at tags, so we avoid doing the user-data lookup when the cache is dropped...
Wasn't this what we agreed would make your development setup not be slow?


Overall, I really like this... I think the end outcome here is much better than keyName. Using tags and spotRequestId + a cache that we try to persist seems robust.

jhford:
Yah, probably.  I forgot that TagSet was returned in all describe* calls, so it's basically free!  please take a look when you wake up today :)

src/aws-manager.js
```diff
@@ -139,10 +125,221 @@ class AwsManager {
...
+              owned = true;
+            }
+            break;
+          case 'Name':
+            workerType = tag.Value;
+            break;
+          default:
+            break;
+        }
+      }
```
jonasfj: You really should start to use lodash for things like this.... This would be:
```js
let workerType = (_.find(resource.TagSet, {Key: 'Name'}) || {}).Value;
```

Really, that's it... Less code = less bugs. And it's a lot faster to read...
(sure, you have to look up the lodash docs to check that _.find() will work like this. But when reading the code the intend is much clearer without much analysis.

jonasfj:
r+

jhford:
f8b202b34f4e3709f519a478d97fc6eb8990b281 deployment had these changes

PR #99, opened 2016-08-01
A different attempt on bug 1264956

jonasfj:
This is just a quick hack to illustrate what I mean...
@jhford, I haven't tested this... It's just an idea... based on a quick review of:
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Response.html#error-property

Maybe, we should only filter away for the given region/workerType... As the workerType might work in other regions...

### DIFF
--- a/src/provision.js
+++ b/src/provision.js
@@ -310,6 +310,13 @@ class Provisioner {
         debug('spawned a %s', toSpawn.workerType.workerType);
         await d();
       } catch (err) {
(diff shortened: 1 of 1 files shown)


jonasfj:
Note: this is more of a comment for: https://github.com/taskcluster/aws-provisioner/pull/95

Rather than an actual PR... feel free to take inspiration and just close this :)

jhford:
lets focus on #106 instead

PR #100, opened 2016-08-03
Limit by time instead of requests

jonasfj:
Instead of limiting to 400 requests... We limit to 3 minutes of provisioning...

Perhaps we need to also reduce the time we sleep between iterations to be a minimum time between iteration starts.. Instead of an absolute time between iterations...
So after each iteration we do `sleep(Math.max(min_iteration_time - iteration_time, 10 * 1000))` where `min_iteration_time = 75 * 1000` is the minimum time between start of an iteration and next start of an iteration...

@jhford, this is more of a suggestion than a PR... Just intended to illustrate an idea.
Up to you if we try it out or not...

### DIFF
--- a/src/provision.js
+++ b/src/provision.js
@@ -13,6 +13,7 @@ let series = require('./influx-series');
 
 const MAX_PROVISION_ITERATION = 1000 * 60 * 10; // 10 minutes
 const MAX_FAILURES = 15;
+const MAX_PROVISIONING_TIME = 3 * 60 * 1000; // 3 minutes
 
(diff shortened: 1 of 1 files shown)


jonasfj:
Note the logic behind the idea is that if we've been provisioning for 3 minutes we should re-evaluate the state again before we continue provisioning... Because it's fairly likely that workers will have processed some of the pending tasks, and likely that new tasks may be pending for a workerType that didn't have pending tasks before.

Note: I suspect we have to change the 75s sleep between iterations to be a minimum sleep time instead...

PR #101, opened 2016-08-15
Submit spot requests by region

jhford:
Let's do each region on its own.


jhford:
Note that this has a lot of #98 in it, it's meant to be landed on top of that.

src/provision.js
```diff
@@ -285,42 +285,51 @@ class Provisioner {
...
-    forSpawning = forSpawning.slice(0, 400);
+    let byRegion = {};
+    for (let x of forSpawning) {
+      assert(x.bid && x.bid.region);
+      let r = x.bid.region;
+      if (!byRegion[r]) {
+        byRegion[r] = [x];
+      } else {
+        byRegion[r].push(x);
+      }
```
jonasfj: You don't need this much code... or a mapping from region to requests...
You just need:
```js
let regions = _.uniq(_.map(forSpawning, ['bid', 'region']));
regions.map(region => _.filter(forSpawning, [['bid', 'region'], region]));

await Promise.all(regions.map(async (forSpawning) => {
  forSpawning = forSpawning.slice(0, 400);
  for (let toSpawn of inRegion) {
  ....
}));
```

jonasfj:
# `r+`

jhford:
f8b202b34f4e3709f519a478d97fc6eb8990b281 deployed this

PR #102, opened 2016-08-16
do automated retries

jhford:
Basically, this makes requestSpotInstance calls idempotent, which makes them retryable.  This means that we can use the in-built retry functionality of the AWS-SDK library.  On my work desktop talking to eu-central-1, after the initial burst of 0-retry sucesses < 0.3s, I start getting 0.5-1.5s requests and >0 retries.

this works pretty well so far.

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -8,6 +8,7 @@ let series = require('./influx-series');
 let _ = require('lodash');
 let delayer = require('./delayer');
 let amiExists = require('./check-for-ami');
+let slugid = require('slugid');
 
 const MAX_ITERATIONS_FOR_STATE_RESOLUTION = 20;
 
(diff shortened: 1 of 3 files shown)


src/aws-manager.js
```diff
@@ -1292,21 +1293,34 @@ class AwsManager {
...
+    log.debug('requesting spot instance');
+    try {
+      spotRequest = await this.ec2[bid.region].requestSpotInstances({
+        InstanceCount: 1,
+        Type: 'one-time',
+        LaunchSpecification: launchInfo.launchSpec,
+        SpotPrice: bid.price.toString(),
+        ClientToken: clientToken,
+      }).promise();
+      console.log('hi');
```
jonasfj: remove this...

jonasfj:
# `r+`

YAY, this is a simple thing that we should have done a long time ago...

PR #103, opened 2016-08-16
make stuff gooder

jhford:


jonasfj:
# `r+`

src/provision.js
```diff
@@ -14,6 +14,21 @@ let series = require('./influx-series');
...
+ * the WorkerType entity, the name is input[n].workerType.workerType.  The
+ * second item is a bid.  The bids have properties region, type (InstanceType),
+ * zone and some pricing information.
+ *
+ * Do not edit the items in the list, just copy them into a new list.  Treat
+ * them as immutable.
+ */
+function orderThingsInRegion(input) {
+  return input;
+}
```
jonasfj: I think this is for dustin to play with :)

jhford: @imbstack also expressed an interest in that as well.

djmitche: #105 :)

PR #105, opened 2016-08-16
Bug 1295673: orderThingsInRegion by round-robin

djmitche:


djmitche:
I think #103 got backed out along with the rest here, so this patch no longer applies (nothing calls orderThingsInRegion).

djmitche:
Rebased on top of #106

PR #106, opened 2016-08-16
saved work from the key pairs branch

jhford:
This makes the improvements around how we submit spot instance requests.  This includes submitting per region as well as using backoffs and idempotent requests to the ec2 api


jhford:
thankfully there weren't many diffhunks that were touched by both patches.  This is working locally.  If you can r? tonight I'll deploy tomorrow.

PR #107, opened 2016-08-17
Add lib-monitor

imbstack:
This also makes includes changes I needed to get the tests run for me (including making some skipped). Not sure if those changes are desirable or reasonable at all.

I'm happy to make whatever changes you see fit for this. This is mostly me guessing at how this might work for the time being.

Forgive anything crazy in here. It's quite late and I'm not quite sure my head is still working.

I didn't know travis was turned back on in this project and it seems this branch isn't working there quite yet. I think I probably need to set some of my user-config.yml stuff in the environment there, but I'll do that tomorrow morning.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -108,8 +108,8 @@ test:
     allowedRegions:
       - 'us-west-2'
   server:
-    publicUrl: https://localhost:60407
-    port: !env:number PORT
+    publicUrl: http://localhost:60407
+    port: 60407
   azure:
     account: jungle
   deadmanssnitch:
--- a/config.yml
+++ b/config.yml
@@ -1,3 +1,5 @@
+---
+
 defaults:
   app:
     publishMetaData: !env:bool PUBLISH_METADATA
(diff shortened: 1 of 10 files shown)


jhford:
Cool!  I like the apporach of having a single call site for the influx and lib-monitor stuff, that's excellent.  In the long run once influx is gone, i'd like to just call monitor.measure directly in the thing if you don't object.

I hope you don't mind, but I'm going to stack a couple more commits into the branch, let me know if you disagree with them and I'll take them out.

Travis looks to be missing the blob storage jungle creds.

jhford:
Oh, also to rebase onto current master after today's landing :)

jhford:
I like this and will merge it.  I have a couple things that I'd like to change a little bit, so I'll open a PR into your monitoring branch.  If you agree with the changes, we can merge it to here then here to master.

jhford:
this was merged, not sure why github didn't reflect that.

PR #108, opened 2016-08-17
extra metrics to give a more overall view

jhford:
I sort of want to have an overview of a bunch of these metrics.  My understanding of statsum is that it's impossible to combine the value of two different measurements without having the internal datastructures that statsum itself has.  Let me know if there's a better way to do this!

### DIFF
--- a/src/monitors.js
+++ b/src/monitors.js
@@ -13,7 +13,15 @@ module.exports.lag = (monitor, series, provisionerId, region, az,
     didShow,
     lag,
   });
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.lag`, lag);
+  monitor.measure(`${region}.lag`, lag);
+  monitor.measure(`${region}.${az}.${instanceType}.lag`, lag);
+  if (didShow) {
+    monitor.count(`overall.didshow`, 1);
+    monitor.count(`${region}.${az}.${instanceType}.didshow`, 1);
+  } else {
+    monitor.count(`overall.noshow`, 1);
+    monitor.count(`${region}.${az}.${instanceType}.noshow`, 1);
+  }
 };
 
 module.exports.spotRequestSubmitted = (monitor, series, provisionerId,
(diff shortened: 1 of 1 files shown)


src/monitors.js
```diff
@@ -13,7 +13,15 @@ module.exports.lag = (monitor, series, provisionerId, region, az,
     didShow,
     lag,
   });
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.lag`, lag);
+  monitor.measure(`${region}.lag`, lag);
+  monitor.measure(`${region}.${az}.${instanceType}.lag`, lag);
```
jonasfj: I would record this both with and without region+az dimensions...

src/monitors.js
```diff
@@ -29,9 +37,13 @@ module.exports.spotRequestSubmitted = (monitor, series, provisionerId,
...
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.spot.bid`, bid.price);
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.spot.price`, bid.truePrice);
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.spot.bias`, bid.bias);
+  monitor.measure('overall.price-per-capacity-unit', bid.truePrice);
+  monitor.measure(`${region}.${az}.${instanceType}.overall.spot.bid`, bid.price);
+  monitor.measure(`${region}.${az}.${instanceType}.overall.spot.price`, bid.truePrice);
+  monitor.measure(`${region}.${az}.${instanceType}.overall.spot.bias`, bid.bias);
+  monitor.measure(`${region}.${az}.${instanceType}.worker.${workerType}.spot.bid`, bid.price);
+  monitor.measure(`${region}.${az}.${instanceType}.worker.${workerType}.spot.price`, bid.truePrice);
+  monitor.measure(`${region}.${az}.${instanceType}.worker.${workerType}.spot.bias`, bid.bias);
```
jonasfj: Consider not adding the region, or submitting both with and without region/az

Often times it's easier to look at fewer dimensions... And then explore deeper...

src/monitors.js
```diff
@@ -46,7 +58,12 @@ module.exports.spotRequestFulfilled = (monitor, series, provisionerId, region, a
     instanceId,
     time,
   });
-  monitor.measure(`${region}.${az}.${instanceType}.${workerType}.spot.filled`, time);
+  monitor.count('overall.spot.died.count', 1);
+  monitor.measure('overall.spot.died.time', time);
+  monitor.count(`${region}.${az}.${instanceType}.overall.spot.filled.count`, 1);
+  monitor.measure(`${region}.${az}.${instanceType}.overall.spot.filled.time`, time);
+  monitor.count(`${region}.${az}.${instanceType}.worker.${workerType}.spot.filled.count`, 1);
+  monitor.measure(`${region}.${az}.${instanceType}.worker.${workerType}.spot.filled.time`, time);
```
jonasfj: Again, why not both with and without?

jonasfj: Particularly I suspect many of the metrics are interesting for a instance type across regions...

jonasfj:
Looks good to me... If you in a hurry land it... Otherwise, let bstack have a look too :)

imbstack:
r+

PR #112, opened 2016-08-22
22 August changes.

jhford:
Here are a bunch of changes that make stuff better.  The first commit is pretty basic, just an update of dependencies.

ddbda88 is a simple patch.  the `AWS.Request.promise()` method is now baked into the upstream client and they even mention it being the supported mechanism.  It now returns the exact same thing as our wrapper, with a `.data` propery on the object.  There's no reason to use our own wrapper any further.

6152087, 2dff3cd is the result of some manual testing.  I wanted to double check that all of this code worked, turns out it didn't.  I fixed all of the issues in there and now it works.

c1a435b is the motivator for the two preceeding patches.  This is a significantly stronger validity check for worker type definitions.  I think we could definately do some more testing of things in this method.  I would prefer to do this over just submitting the spot instances and having them fail.  It's a pretty quick check to see if a worker type is valid before and we can skip the case where there are e.g. 10000 requests for a worker type that's invalid but there's pending work for actually valid worker types.  Let's not waste any time on things we know to be invalid.  There's also intelligence done here to ensure that we only ever check each worker type at most once for a given iteration.


config.yml
```diff
@@ -33,6 +33,9 @@ defaults:
     forceSSL: false
     trustProxy: false
 
+  monitor:
+    project: 'aws-provisioner-dev'
```
jonasfj: Use `mock: true` for monitor when testing..

jhford: that's already done in `src/main.js` by checking for `profile === 'test'`, but i suspect what we really want is `profile !== 'production'`

jhford: `[changes-22-8-2016 `6f5d2a2`] make the choice of mock or not mock explicit`

src/provision.js
```diff
@@ -100,6 +120,40 @@ class Provisioner {
...
 
+  async __filterBrokenWorkers(workerTypes, forSpawning) {
+    assert(workerTypes);
+    assert(Array.isArray(workerTypes));
+    assert(forSpawning);
+    assert(Array.isArray(forSpawning));
+    // Let's find the unique worker types in the list of requests to make This
+    // is to avoid wasting time checking on worker types which we aren't even
+    // going to make an attempt on
+    let toSpawnWorkerTypes = _.intersection(forSpawning.map(x => x.workerType.workerType));
```
jonasfj: I think you're looking for `_.uniq(` are you not?

src/provision.js
```diff
@@ -100,6 +120,40 @@ class Provisioner {
...
+    if (disabled.length > 0) {
+      let err = new Error('Found invalid worker types');
+      err.types = disabled;
+      //this.monitor.reportError(err);
+      log.warn({disabled}, 'found worker types which cannot launch, ignoring them');
+    }
+
+    console.log(disabled);
+
+    return forSpawning.filter(x => disabled.indexOf(x.workerType.workerType) === -1);
```
jonasfj: how about using `_.includes` It reads much easier...

src/provision.js
```diff
@@ -100,6 +120,40 @@ class Provisioner {
...
+    }
+
+    if (disabled.length > 0) {
+      let err = new Error('Found invalid worker types');
+      err.types = disabled;
+      //this.monitor.reportError(err);
+      log.warn({disabled}, 'found worker types which cannot launch, ignoring them');
+    }
+
+    console.log(disabled);
```
jonasfj: Intentional?

src/provision.js
jonasfj: same above...

jonasfj:
# `r+`

Please, consider using `_.includes` and `_.uniq`

jhford:
@jonasfj Fwiw,
```
> require('lodash').intersection([1,1,1,1,1,2,3,3,1,2,3,2,1,4,5])
[ 1, 2, 3, 4, 5 ]
> require('lodash').uniq([1,1,1,1,1,2,3,3,1,2,3,2,1,4,5])
[ 1, 2, 3, 4, 5 ]
>
```
I know that `_.uniq` describes the intent better, but both do work in this specific case... not saying that I should continue using `_.intersection()`

jhford:
maybe we could call `amiExists()` something like `describeImage`... I'm not opposed to reusing EC2 api method names, and already do that a bit.

PR #113, opened 2016-08-22
Add AMIs validation to AmiSets endpoints

anarute:
Create `Amiset.validate()` function that is called on create and update endpoints and will be used before applying an AmiSet to a WorkerType in the AmiSet management tool. 
This function checks if all AMIs from a given AmiSet are valid and if they are stored with the right virtualization type - this validation is done using the new method called `getVirtualizationType()`.


test-src/manageamiset_test.js
```diff
@@ -85,6 +87,11 @@ describe('AMI Set api', () => {
     });
   });
 
+  it('should be able to check if the AMIs from an AMI set are valid', async () => {
+    await client.createAmiSet(id, amiSetDefinition);
+    assume(await client.validateAmiSet(id)).to.deeply.equal([]);
```
anarute: I wanted to test a known invalid amiSet here, but I didn't knkow how to do it since with the validation method we cannot create "fake" amiSets anymore.
I tried to do something like `Amiset.validate(context, fakeAmiSet)`, but I don't know what to pass as `context`.

anarute:
btw @jhford, I can only test this if I use dev table names without this "2"s at the end: https://github.com/taskcluster/aws-provisioner/blob/master/config.yml#L9
I don't know if that's a problem, but locally my config is like this:

    workerTypeTableName: DevWorkerTypes
    amiSetTableName: DevAmiSets
    secretTableName: DevWorkerSecrets
    workerStateTableName: DevWorkerState

jhford:
@anarute regarding table names, I only put a 2 on there because someone else had modified the Entity that was on that table.  If you're able to use the non-2 suffixed tables, great.

src/get-virtualization-type.js
```diff
@@ -0,0 +1,39 @@
+let log = require('./log');
+let debug = log.debugCompat('aws-provisioner:lib:check-for-ami');
```
jhford: Let's not use log.debugCompat and instead use the log methods directly.

Here's some things you can do:
```javascript
log.trace({a:1}, "this is so unimportant we likely never show it");
log.debug("can log with just a string");
log.debug({b: 2}, "or also with an object.  objects let us inspect logs better");
log.info("the next log level up");
log.warn("and another level up");
log.err("here's were the good stuff is");
let anError = new Error('Sample Error');
log.error(anError, "This will do some *really* nice things like logging full stacks");
log.error({err: anError, otherField: 123}, "You can also make it the 'err' property of the object");
log.error("or just ignore");
log.fatal("this is basically just when the program is about to quit");
```

If there are time concerns around the internship timing, this is really not a huge deal, but it would be nice.  The string should always be static (no %s, %j or +'ing strings together.  The object should have enough information to debug.  Usually, this means including the region and ami set or something.

src/get-virtualization-type.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (result.data.Images.length === 0) {
+    throw new Error('Image does not exist');
+  } else if (result.data.Images.length > 1) {
+    debug(result.data);
+    throw new Error('Image returned more than one result');
+  }
+
+  if (result.data.Images[0].ImageId === ami) {
+    return result.data.Images[0].VirtualizationType;
```
jhford: This file is nearly identical to `ami-exists.js`.  I think we should consider merging them in that case.  In the places that I'm testing just for AMI existance, a return value of 
```javascript
return result.data.Images[0]
```
would pass the truthy or falsy questions I have of the answer to whether an ami exists, so we could probably merge the code into the same branch.  It would be good to rebase onto the `changes-22-8-2016` branch as that's going to be landing first thing in the Berlin.

test-src/manageamiset_test.js
jhford: What I usually end up doing is something like this: https://github.com/taskcluster/aws-provisioner/blob/master/test-src/badworkertype_test.js#L69

There's a slight issue in the implementation of this test, it shouldn't be throwing in the try block, but do you get the idea?  basically we handle the error and if the error is one that we don't expect, we re-throw to mark as a failure.

src/ami-set.js
```diff
@@ -51,6 +55,49 @@ AmiSet.listAmiSets = async function () {
 };
 
 /**
+ * Checks if the AMIs from the amiSet are valid or not.
+ */
+
+async function checkAmi(ctx, region, ami, vtype) {
```
jhford: So, given that you're passing in the scope anyway through `ctx`, let's make this an extra `AmiSet` method.  `AmiSet.__checkAmi(region, ami, vtype)` would be great.

src/api-v1.js
```diff
@@ -667,11 +702,19 @@ api.declare({
   }
 
   let loadedAmiSet = await this.AmiSet.load({id});
+  let invalidAmis = await this.AmiSet.validate(this, loadedAmiSet);
```
jhford: Let's validate before we load... We know that if we don't validate that we can return immediately.  Chances are also good that we're more likely to fail validation than loading, so not too much risk in wasted effort.

src/get-virtualization-type.js
anarute: Yes, I was afraid to merge them and mess up where ami-exists was been used. This solution sounds good for me! How can we merge them? should I send a different PR above `changes-22-8-2016` only with this change?

src/ami-set.js
jhford: Actually, I see why you're doing it now.  It's similar to the `createLaunchSpec` method, since we have to have a non-instance method for doing the check before creating and modifying, and Constructor methods as I suggested wouldn't have access to context.  I think that doing the following would match what we have elsewhere:

```javascript
//Helper method
AmiSet.__checkAmi = function (region, ami, vtype, ec2) { } 

// A constructor method available
AmiSet.validate = function (region, ami, vtype, ec2) { } 

// An instance copy
AmiSet.prototype.validate = function (region, ami, vtype) {
  return AmiSet.validate(region, ami, vtype, this.ec2);
} 

```

src/ami-set.js
```diff
@@ -51,6 +55,49 @@ AmiSet.listAmiSets = async function () {
...
+  let request;
+  let missing = [];
+  let virtualizationType = '';
+
+  if (ami) {
+    request = {
+      ImageIds: [ami],
+    };
+    let exists = await amiExists(ctx.ec2[region], ami);
```
jhford: let's generalise this instead of doing the same describeImages call twice.  No reason to call it once to see if it's there, then again to get more info.  If you make the changes that I suggested to ami-exists.js then this would be a little easier.

src/ami-set.js
```diff
@@ -51,6 +55,49 @@ AmiSet.listAmiSets = async function () {
...
 /**
+ * Checks if the AMIs from the amiSet are valid or not.
+ */
+
+async function checkAmi(ctx, region, ami, vtype) {
+  let request;
+  let missing = [];
+  let virtualizationType = '';
+
+  if (ami) {
```
jhford: Hmm.  this is likely to make things silently fail.  I assume this is to check for the case that the hvm/pv thing is set to '' or not defined.  Let's instead not even call this function if the value is undefined.  In fact, let's have the first three lines of this function be assertions that the parameters passed in are as expected.  This makes debugging later easier.
```javascript
assert(typeof region === 'string');
assert(typeof ami === 'string');
assert(typeof vtype === 'string');
```

src/ami-set.js
```diff
@@ -51,6 +55,49 @@ AmiSet.listAmiSets = async function () {
...
+  return missing;
+};
+
+AmiSet.validate = async function (ctx, amiSet) {
+  let missing = [];
+  let exists = false;
+
+  await Promise.all(amiSet.amis.map(async (def) => {
+
+    if (def.hvm) {
```
jhford: ahh you are already only calling if it is defined.  No reason for the `if (ami)` in your helper method then, right?

src/ami-set.js
```diff
@@ -51,6 +55,49 @@ AmiSet.listAmiSets = async function () {
...
+      missing = missing.concat(await checkAmi(ctx, def.region, def.hvm, 'hvm'));
+    }
+    if (def.pv) {
+      missing = missing.concat(await checkAmi(ctx, def.region, def.pv, 'paravirtual'));
+    }
+  }));
+
+  return missing;
```
jhford: Maybe return an object like this:
```javascript
return {
  valid: valid,
  invalidAmis: missing
};

```
To track the valid, what you could do is something like `let valid = true` and then only set it to false when you have a condition that invalidates validity.

As well, instead of returning the list of missing ones, let's return a more general set of information.  Missing implies that the AMI is invalid, but what if it fails because of the wrong virtualization type instead of abesnce?  For the invalidAmis, a list of objects like `{region: ..., ami: ..., vtype: ...}` would be really helpful.

src/get-virtualization-type.js
jhford: You could learn how to do a rebase!  (not because I'm lazy, but it's good experience).

Here's what I'd do, roughly.
```bash
git checkout -t <mygoodlocalbranch> -b rebase-on-changes-22-8-2016
git rebase origin/changes-22-8-2016
```
If it doesn't print any errors, do `git diff origin/changes-22-8-2016` to verify things worked as you expected.  If it does have errors you can fix them up with this: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/

anarute: Got it! I misunderstood what you said in the first comment, I'll do that. 
I'm already familiar with rebase, thanks for the link and tips anyway :)

src/ami-set.js
anarute: I already return the vtype because I need to check that in the amiset tool: `missing = missing.concat(await checkAmi(ctx, def.region, def.hvm, 'hvm'));`
Now, adding the boolean `valid` on it too, it's much better indeed.

src/check-for-ami.js
```diff
@@ -39,7 +39,7 @@ module.exports = async function (ec2, ami) {
   }
 
   if (result.data.Images[0].ImageId === ami) {
-    return true;
+    return result.data.Images[0];
```
anarute: @jhford as we discussed, I changed what `check-for-ami` method returns, apparently it didn't break anything outside amiset, but it would be great if you could check it too, please :)

anarute:
@jhford the changes I made:

- deleted `get-virtualization-type.js` and changed what `check-ami` returns, so we don't need to call `describeImage` twice.
- make `AmiSet.__checkAmi` a helper function
- add `AmiSet.prototype.validate` function
- remove `if` condition from `AmiSet.__checkAmi` since we already check the ami ids before calling it
- add test to validate function

djmitche:
@jhford will you get a chance to look at this once more?

src/check-for-ami.js
jhford: yep, this should be fine.  Going from `true` to `Object` should pass all the same truthiness checks that exist elsewhere in the codebase

jhford:
I merged this prematurely, and github is not showing me the squash and merge thing any longer.

jhford:
why won't github show this as opened any longer :(

walac:
@jhford once merged, it closes the PR, you cannot reopen it again..

anarute:
@jhford is there anything I can do to help?

jhford:
github :(

@anarute if you could do a rebase locally, which should be a no-conflict one (git fetch origin && git rebase origin/master) then open a new PR, we could use that :)

walac:
I am afraid when @anarute does that, she will get the same commits set as the upstream, if I am following this right, and there will be nothing to open a PR for...

PR #114, opened 2016-08-23
Include internal state provisioner

jhford:
This includes c772d4d which is a fix to the fallout from today's deployment's issue.

The rest of this patch is to include the internally tracked EC2 spot instance requests in the provisioner state apis.  The new parameter `visibleToEC2Api` will tell us whether the instance is included in the crappy API lag situation we get.

### DIFF
--- a/src/api-v1.js
+++ b/src/api-v1.js
@@ -92,10 +92,6 @@ function workerTypeSummary(workerType, workerState) {
     } // note that other states are ignored
   });
 
-  workerState.requests.forEach(request => {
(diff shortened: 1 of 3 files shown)


jonasfj:
# `r+`

jhford:
https://github.com/taskcluster/taskcluster-tools/pull/132 see also

jonasfj:
This broke the UI...

jonasfj:
`internalTrackedRequests` is missing

PR #116, opened 2016-08-24
Check that all requested security groups are present in the region requested for.

jhford:
This is one of the last things that I know of that a DryRun might not catch.  The only other thing that I suspect is possible to cause this issue is the KernelID, but we don't yet specify that level of detail.  I've tested this code pretty well and it works.  I even wrote some unit tests!

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -8,6 +8,7 @@ let monitors = require('./monitors');
 let _ = require('lodash');
 let delayer = require('./delayer');
 let amiExists = require('./check-for-ami');
+let sgExists = require('./describe-security-group');
 let slugid = require('slugid');
 
 const MAX_ITERATIONS_FOR_STATE_RESOLUTION = 20;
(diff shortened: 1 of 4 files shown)


src/aws-manager.js
```diff
@@ -824,9 +825,18 @@ class AwsManager {
...
       await Promise.all(worker.instanceTypes.map(async t => {
         let launchSpec = launchSpecs[r.region][t.instanceType].launchSpec;
 
+        if (launchSpec.SecurityGroups) {
+          for (let sg of launchSpec.SecurityGroups) {
+            if (!_.includes(allSGForRegion, sg)) {
+              allSGForRegion.push(sg);
+            }
+          }
+        }
```
jonasfj: Don't mix multiple things into the same loop... This is hard to read...
```js
let launchSpecsRegion = worker.instanceTypes.map(t => launchSpecs[r.region][t.instanceType].launchSpec);
let allSGForRegion = _.uniq(_.flatten(launchSpecsRegion.map(spec => spec.SecurityGroups)));
```

jonasfj:
# `r+`

But please untangling some of the loops...

src/aws-manager.js
jhford: fair.  it was an admittedly poor attempt at optimizing by not going through the list twice. :)

jonasfj: I know I catch myself doing the same thing many times...

I've learn in JS that if it's not doing I/O, I should keep it simple as oppose to trying and be smart.

jhford: i know, i usually do that.  On second thought, mutating things outside of a map in the map's callback is a bad idea just in general :)

let me agree more aggressively!

PR #117, opened 2016-08-24
Trace logging for incorrect pending capacity

jhford:
Here's some extra logging for this.  It's going to generate a decent amount of extra output, but I don't think that's a huge concern.  I'm also going to add in something to print out when we get into this weird state.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 web: node lib/main server | ./node_modules/.bin/bunyan -o short -l info
-provisioner: node lib/main provisioner | ./node_modules/.bin/bunyan -o short -c "this.level > 20 || this.capacityForTypeLog"
+provisioner: node lib/main provisioner | ./node_modules/.bin/bunyan -o short -l info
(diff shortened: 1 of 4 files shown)


jhford:
@djmitche Do you think this is enough extra logging to tell us what's going on?  My brain is a little fried.  Since this isn't causing tree closure right now, I'm going to leave it for tonight and include this with the deployment of the check-for-security-groups branch tomorrow.

jhford:
Example of the capacityForType trace stuff:
```
22:30:56.333Z  INFO aws-provisioner-development: capacityForType trace (answer=1, workerType=cli)
    states: [
      "running"
    ]
    --
    trace: [
      {
        "newCapacity": 1,
        "capacityOfType": 1,
        "instanceType": "r3.xlarge",
        "region": "us-west-1",
        "state": "running",
        "type": "instance"
      }
    ]
```
each item in the array is a thing we counted against.  This is going to be kind of crazy for production logs :/

jhford:
This is the changeForType outcome message example
```
22:30:56.334Z  INFO aws-provisioner-development: changeForType outcome (workerType=cli, pendingTasks=0, runningCapacity=1, pendingCapacity=0, spotReqCap=0, pendingInstCapacity=0, change=0)
```

jhford:
Once this is resolved, I'm going to turn ` if (pendingCapacity < 0 && -pendingCapacity > pendingTasks)` into an assertion.  That case should never ever happen.

jhford:
This is live in production

PR #119, opened 2016-08-25
bug 1298010 make the last modified time of a worker type defintion available without auth

jhford:


jhford:
```
$ curl -i http://localhost:5557/v1/worker-type-last-modified/cli
HTTP/1.1 200 OK
X-Powered-By: Express
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: OPTIONS,GET,HEAD,POST,PUT,DELETE,TRACE,CONNECT
Access-Control-Request-Method: *
Access-Control-Allow-Headers: X-Requested-With,Content-Type,Authorization,Accept,Origin
Content-Type: application/json; charset=utf-8
Content-Length: 71
ETag: W/"47-301c08ca"
Date: Thu, 25 Aug 2016 13:30:19 GMT
Connection: keep-alive

{
  "workerType": "cli",
  "lastModified": "2016-08-25T11:46:15.999Z"
}
```

src/worker-type.js
```diff
@@ -499,6 +499,7 @@ WorkerType.createLaunchSpec = function(bid, worker, keyPrefix, provisionerId, pr
     price: bid.truePrice,
     launchSpecGenerated: new Date().toISOString(),
     workerModified: worker.lastModified.toISOString(),
+    lastModified: worker.lastModified.toISOString(),
```
jhford: This is because I'm not sure if anyone was using workerModified and it's not that much extra space.

src/api-v1.js
```diff
@@ -321,6 +321,45 @@ api.declare({
...
+  let workerType = req.params.workerType;
+
+  let worker;
+  try {
+    worker = await this.WorkerType.load({workerType: workerType});
+
+    // We do this because John made a mistake in the V1->V2
+    // schema update and there was a typo :(
+    let workerjson = worker.json();
+    return res.reply(_.pick(workerjson, 'workerType', 'lastModified'));
```
petemoore: Would this also work with:
```
return res.reply(_.pick(workerjson, 'workerType', 'workerModified'));
```
and then not add `lastModified` to `worker-type.js`?

jhford: Nope, because that property isn't defined in the worker type definition.  I could acheive what you're suggesting by doing:
```javascript
return res.reply({
  workerType: workerjson.workerType,
  workerModified: workerjson.lastModified,
});
```
but I chose not to because I'd like to make the return of this function a pure subset of the full worker type definition, rather eventually removing the `workerModified` field at some point in the future.

@gregarndt do you know if you're using `UserData.workerModified` in the docker-worker?

gregarndt: We're not using it.

petemoore:
## r+ ##

Also ok with me if you remove the `UserData.workerModified` at the same time.

PR #121, opened 2016-08-26
Be more explicit with capacity information, the fallback was causing issues

jhford:
Stop having a fallback for the case that we can't figure out capacity for a worker type.

Also rewrite manage.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "babel-eslint": "^6.0.0",
     "babel-preset-taskcluster": "^3.0.0",
     "babel-runtime": "^6.6.1",
+    "cli-table": "^0.3.1",
     "commander": "^2.6.0",
     "debug": "^2.2.0",
     "eslint": "^2.10.2",
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -906,97 +906,59 @@ class AwsManager {
 
     for (let instance of instances) {
       if (_.includes(states, instance.State.Name)) {
-        try {
-          capacity += workerType.capacityOfType(instance.InstanceType);
+        let instanceType = instance.InstanceType;
+        assert(typeof instanceType === 'string');
+        capacity += workerType.capacityOfType(instanceType);
 
-          capacityTrace.push({
-            newCapacity: capacity,
-            capacityOfType: workerType.capacityOfType(instance.InstanceType),
-            instanceType: instance.InstanceType,
-            region: instance.Region,
-            state: instance.State.Name,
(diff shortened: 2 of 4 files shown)


djmitche:
It's hard to tell in the midst of the rewrite and issues-fixing what this changes.  How does it address bug 1297811?

jhford:
manage.js being rewritten doesn't have anything to do with the bug.   	f197ab7 and 2adcf30 are the commits that matter here.

I suspect that part of the cause of this bug is that because of the eventual consistency, we have instances that show up in the spot requests list with a different amount of lag than the corresponding instance in the instances list.  I don't know if that accounts for being 150 off, but it wouldn't surprise.

jonasfj:
This looks okay to me. I didn't really look at manage.js since it's just a utility how much damage can it do :)

Note: as previously advocated `_.map` and `_.flatten` are better than nested for-loops. But in this case the code is not that hard to read..

Perhaps something to make this simple is whenever you define a list of some sort of objects, write a comment like:
```js
// ignoreSR is spot-requests that have been fulfilled and shouldn't be considered stalled
// These are objects on the following form:
// {
//   Region: 'us-west-2',
//   WorkerType: '...',
//   KeyName: '...',
//   TODO: List all the properties that we rely on the EC2 API to return.
// }
let ignoreSR = [];
```

A crazy thing might even be to define a JSON schema for the objects and validate them against that.
But just having a comment for the many lists of state outline what properties must be there and what they are would be really useful. And we would be able to write a simple `validateState` function that ensures these invariants hold.

This would be very useful when trying to write new code too :) Even if some of the properties we document like this are also documented in the EC2 api docs. Note: docs like this can largely just be an example...

jonasfj:
Note: Maybe the above thing about documenting state structure is irrelevant to this PR.. But just something to consider.

djmitche:
Perhaps off-topic, but I really like the idea.  Most of my time reading AWS provisioner source is spent chasing down where a particular data structure came from and what it looks like.

djmitche:
Thanks for the explanation.  It is hard to see how we had 300+ spot requests for gecko-decision, but this could certainly be *part* of that, as you say.  r+ :)

jhford:
if we're seeing silly amounts of spot requests, then it's the API returning silly amounts of spot requests...  There's not much we can do about the api flat out lying.

djmitche:
That may be, but I'd like to see evidence of it (that would require it to invent srids, I think, so it seems unlikely)

PR #122, opened 2016-08-26
make fewer EC2 API Calls

jhford:
@jonasfj i know you've wanted this for a while, but I think it's time.  I don't want to merge the in-process lists because I still think they should be seperate so we don't have to `.filter()` literally every usage.

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -211,7 +211,7 @@ class AwsManager {
               },
               {
                 Name: 'instance-state-name',
-                Values: ['running', 'pending'],
+                Values: ['running', 'pending', 'shutting-down', 'terminated', 'stopping'],
               },
             ],
           }).promise(),
@@ -223,32 +223,7 @@ class AwsManager {
                 Values: [this.keyPrefix + '*'],
               }, {
                 Name: 'state',
-                Values: ['open'],
-              },
-            ],
-          }).promise(),
-          // Dead instances
-          ec2.describeInstances({
-            Filters: [
-              {
-                Name: 'key-name',
-                Values: [this.keyPrefix + '*'],
-              },
-              {
-                Name: 'instance-state-name',
-                Values: ['shutting-down', 'terminated', 'stopping'],
-              },
-            ],
-          }).promise(),
-          // Dead spot requests
-          ec2.describeSpotInstanceRequests({
-            Filters: [
-              {
-                Name: 'launch.key-name',
-                Values: [this.keyPrefix + '*'],
-              }, {
-                Name: 'state',
-                Values: ['cancelled', 'failed', 'closed', 'active'],
+                Values: ['open', 'cancelled', 'failed', 'closed', 'active'],
               },
             ],
           }).promise(),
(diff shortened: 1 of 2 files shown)


src/aws-manager.js
```diff
@@ -285,7 +260,12 @@ class AwsManager {
...
-            apiState.instances.push(instance);
+            // apiState for instance state name of 'running', 'pending'
+            if (_.includes(['running', 'pending'], instance.State.Name)) {
+              apiState.instances.push(instance);
+            } else if (_.includes(['shutting-down', 'terminated', 'stopping'], instance.State.Name)) {
+              deadState.instances.push(instance);
+            }
             ignoreSR.push(instance.SpotInstanceRequestId);
           }
         };
```
jonasfj: So we could do this in a smarter way:
```js
let instances = _.flatten(response[0].data.Reservations.map(reservation => reservation.Instances.map(instance => {
  instance.Region = region;
  instance.WorkerType = this.parseKeyPairName(instance.KeyName).workerType;
  return instance;
})));
apiState.instances.push(...instance.filter(instance => _.includes(['running', 'pending'], instance.State.Name)));
deadState.instances.push(...instance.filter(instance => _.includes(['shutting-down', 'terminated', 'stopping'], instance.State.Name)));
ignoreSR.push(...instances.map(instance => instance.SpotInstanceRequestId);
```
I think that's about the same...

src/aws-manager.js
```diff
@@ -285,7 +260,12 @@ class AwsManager {
...
             instance.Region = region;
             instance.WorkerType = workerType;
-            apiState.instances.push(instance);
+            // apiState for instance state name of 'running', 'pending'
+            if (_.includes(['running', 'pending'], instance.State.Name)) {
+              apiState.instances.push(instance);
+            } else if (_.includes(['shutting-down', 'terminated', 'stopping'], instance.State.Name)) {
+              deadState.instances.push(instance);
+            }
             ignoreSR.push(instance.SpotInstanceRequestId);
```
jonasfj: hmm... are you not missing some return statements? I'm not following the logic I'll admit... I don't know what `ignoreSR` is good for..

jonasfj:
I would love it if instead of doing:
```js
let response = await Promise.all([
```
we did:
```js
let [instances, spotRequests, availabilityZones, spotPriceHistory] = await Promise.all([
```

Do you don't have to write `response[0]` and `response[1]` and try to remember what they are...

jonasfj:
r+,

It looks like an improvement, but I'll admit I can't keep up with the state tracking logic there too many lists. So I just hope that is still correct :)

jonasfj:
Note: I really like doing fewer EC2 requests, so that's awesome!

src/aws-manager.js
jhford: I think that what I have in the original code is a lot easier to read

src/aws-manager.js
jhford: `ignoreSR` are spot request ids which are for an instance.  If a SRID is associated with an instance, we know that the spot request it's for has be filled.  If an SR has been filled but the describeSpotInstance requests call doesn't reflect that due to eventual consistency, we now know that we can ignore it to avoid double counting it.

The case I'm concenred with for ignoreSR is that the describe calls are out of sync, so we're double counting some spot requests.

jhford:
For the time being, I don't think that I'll have a chance to test this.  Since this patch will likely bitrot, I'm going to close this PR.  If we decide to do this, let's revisit in a new PR

PR #123, opened 2016-08-30
Don't reqiure NODE_ENV to be set just to run tests

djmitche:
The loader ensures that this value is always set when running the
provisioner directly, but during testing it need not be set.

@anarute

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "pretest": "npm run compile",
     "lint": "eslint src/*.js test-src/*.js",
     "lintfix": "eslint --fix src/*.js test-src/*.js",
-    "test": "npm run lint && npm run compile && mocha test/*_test.js"
+    "test": "npm run lint && npm run compile && NODE_ENV=test mocha test/*_test.js"
   },
   "bin": {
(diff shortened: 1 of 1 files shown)


src/log.js
```diff
@@ -1,8 +1,5 @@
 let log = require('taskcluster-lib-log');
 
-let env = process.env.NODE_ENV;
-if (!env) {
-  throw new Error('You must have an environment');
-}
+let env = process.env.NODE_ENV || "test";
```
jhford: test is not the right thing to have as default.
```javascript

let env;
if (!process.env.NODE_ENV) {
  env = ""
} else {
  env = '-' + process.env.NODE_ENV.trim();
}
module.exports = log('aws-provisioner' + env); 
```

I'd rather that we just always set NODE_ENV...

djmitche:
@jhford do you mind taking another look?

jhford:
`NODE_ENV` should be `test` for tests, but otherwise looks good to me.  I have the provisioner in a pipeline now, so feel free to merge with the fix when you choose.

PR #125, opened 2016-09-29
Update ec2 API version

jonasfj:
This caused some issues where we ended up using old default values.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1306167

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -57,7 +57,7 @@ defaults:
     accessKeyId: !env AWS_ACCESS_KEY_ID
     secretAccessKey: !env AWS_SECRET_ACCESS_KEY
     region: us-west-2
-    apiVersion: 2014-01-01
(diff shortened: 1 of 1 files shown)


jhford:
Based on the failures created by this in travis, it looks like there are some changes to how `LaunchSpecifications` are done in the new version of the API.  This is a non-trivial change, and will almost certainly require an audit of the worker-types.  We'll also need to figure out which parts of the launch specifications aren't valid.

jonasfj:
oh, that's depressing.. though probably an important thing to get done.
suboptimal configs are bad when we have 1k nodes.

Den 29. sep. 2016 1.52 AM skrev "John Ford" <notifications@github.com>:

jhford:
Sure, but this is a new requirement, so we need to spend the time to deal with it as we would anything else :)

PR #126, opened 2016-10-02
Add AMIs validation to AMI set

anarute:
New PR for: https://github.com/taskcluster/aws-provisioner/pull/113

cc @jhford @djmitche


djmitche:
Thanks for sticking with this @anarute!

djmitche:
We've elected to remove AMI set support from the provisioner.  I checked in with Ana about this directly.

PR #127, opened 2016-10-21
New ping API

owlishDeveloper:
@imbstack @djmitche r?

### DIFF
--- a/src/exchanges.js
+++ b/src/exchanges.js
@@ -1,7 +1,7 @@
-let base = require('taskcluster-base');
+let Exchanges = require('pulse-publisher');
 
 /** Declaration of exchanges offered by the aws-provisioner */
-let exchanges = new base.Exchanges({
+let exchanges = new Exchanges({
   title: 'AWS Provisioner Pulse Exchanges',
   description: [
     'Exchanges from the provisioner... more docs later',
--- a/src/main.js
+++ b/src/main.js
@@ -6,7 +6,12 @@ let _ = require('lodash');
 let path = require('path');
 
 let taskcluster = require('taskcluster-client');
-let base = require('taskcluster-base');
+let loader = require('taskcluster-lib-loader');
+let config = require('typed-env-config');
+let libMonitor = require('taskcluster-lib-monitor');
+let libValidator = require('taskcluster-lib-validate');
+let libApp = require('taskcluster-lib-app');
+let stats = require('taskcluster-lib-stats');
 
 let workerType = require('./worker-type');
 let secret = require('./secret');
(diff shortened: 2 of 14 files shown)


owlishDeveloper:
Hi @jhford ! I need some help in testing this! I worked on removing all references to taskcluster-base from aws-provisioner. It seems like I didn't miss anything. However, `npm test` yelds two mistakes for me. The reason for the first one seems to be the absence of azureBlob credentials. The second one says:

```
TypeError: Cannot read property 'terminate' of undefined
      at _callee2$ (test-src/helper.js:23:9)
```
I can't figure out what's wrong there.

djmitche:
I'm guessing that the second error is caused by the first.  It comes from https://github.com/taskcluster/aws-provisioner/pull/127/files#diff-4e914f6ee4525908bd15d1153b36f87aL21 where it calls `server.terminate()`, and the server was never set up because of the first error.

djmitche:
Let's see if Travis passes in #129.

djmitche:
No luck - still a bunch of permission errors.

@jhford can you review and test?

src/exchanges.js
```diff
@@ -1,7 +1,7 @@
-let base = require('taskcluster-base');
+let Exchanges      = require('pulse-publisher');
```
owlishDeveloper: Spaces gone!

src/main.js
```diff
@@ -6,7 +6,12 @@ let _ = require('lodash');
 let path = require('path');
 
 let taskcluster = require('taskcluster-client');
-let base = require('taskcluster-base');
+let loader = require('taskcluster-lib-loader');
+let config = require('typed-env-config');
+let Monitor = require('taskcluster-lib-monitor');
```
owlishDeveloper: Done!

src/main.js
```diff
@@ -6,7 +6,12 @@ let _ = require('lodash');
...
+let Validator = require('taskcluster-lib-validate');
```
owlishDeveloper: Done!

src/main.js
```diff
@@ -6,7 +6,12 @@ let _ = require('lodash');
...
+let App = require('taskcluster-lib-app');
```
owlishDeveloper: Done! Thank you for reviewing!

owlishDeveloper:
As for those build errors - I tried to read through https://devcenter.heroku.com/articles/troubleshooting-node-deploys, and I noticed, for example, that we have node 4 in package.json, and I worked on node 5. I doubt this will resolve the problem, but I think it's worth a shot, so I changed node version in package.json.

imbstack:
Perhaps this is arunoda/node-usage#52? If so, it seems to be the node 4 vs node 5 thing.

djmitche:
OK, I think the remaining errors in Travis occur on every PR, so is this good to go? @jhford if so can you land it?

jhford:
This branch doesn't merge correctly into master branch.  I'm going to try to redo the patch based on the diff in this pr :)

owlishDeveloper:
New PR with these changes is here https://github.com/taskcluster/aws-provisioner/pull/135
I'm closing this.

PR #130, opened 2017-02-02
Bug 1335507: sign and encrypt worker types and secrets

djmitche:
With this change, someone intercepting a copy of our table backups will
not get any secrets.  A restore will also verify the signature on the
rows, avoiding the risk of modification.  Otherwise, there is no impact.

Deployment will require setting TABLE_SIGNING_KEY and TABLE_CRYPTO_KEY.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -19,6 +19,12 @@ defaults:
       - 'us-west-2'
       - 'us-east-1'
       - 'eu-central-1'
+
+    # Key for signing in base.Entity (sufficiently random string required)
+    tableSigningKey:          !env TABLE_SIGNING_KEY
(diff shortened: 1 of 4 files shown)


djmitche:
I see those six test failures on master, too, so I think they're unrelated to this push.  I expect running this in staging will be the more realistic test, anyway.

config.yml
```diff
@@ -137,6 +143,9 @@ test:
     awsInstancePubkey: 'fake-pubkey'
     allowedRegions:
       - 'us-west-2'
+    tableSigningKey:          not-a-secret-so-you-cant-guess-it
```
jhford: can you also set a value for the development profile?

PR #133, opened 2017-03-01
state read/write using azure-blob-storage library

elenasolomon:


src/main.js
```diff
@@ -4,6 +4,7 @@ let log = require('./log');
 let aws = require('aws-sdk');
 let _ = require('lodash');
 let path = require('path');
+let fs = require('fs');
```
jhford: Let's use `mz/fs` instead.  It's a module that wraps the entire node standard library in promises.
```
let fs = require('mz/fs');`
```

src/main.js
```diff
@@ -47,6 +49,25 @@ let load = base.loader({
...
+  stateNewContainer: {
+    requires: ['cfg', 'profile'],
+    setup: async ({cfg, profile}) => {
+      let container = `worker-state-${profile}`;
+      let schema = fs.readFileSync(`${__dirname}/../schemas/state-definition.json`, 'utf8');
+      let schemaObj = JSON.parse(schema);
+      let options = {
+        credentials: {
+          accountName: cfg.azureBlob.accountName,
+          accountKey: cfg.azureBlob.accountKey,
```
jhford: If the TC-Auth stuff has landed, can we instead use the accountName from the `cfg.azure` configuration entry and use SAS and TC-Auth?  This will require us to update the taskcluster-client.  If that's not already done, I'll take a look soon.

PR #135, opened 2017-03-03
Start using the new State Storage

jhford:
Here's a few more commits updating the provisioner

@elenasolomon: I copied the unit test for the old `state()` method and got it running in the unit tests.  The code seems to work, except for `provisioner worker type api newState() should return an empty (but not 404) response when no state is available`.  For that test, the following exception is thrown:
```
  azure-blob-storage:utils Failed to load the blob 'VXcI_j3JTZ-drBCnxUp8QA' from container "worker-state-test" with error: ErrorWithoutCodeError: No error message given, in payload '' +482ms
  base:api Error occurred handling: /new-state/VXcI_j3JTZ-drBCnxUp8QA, err: ErrorWithoutCodeError: No error message given, in payload '', as JSON: {"name":"ErrorWithoutCodeError","code":"ErrorWithoutCode","statusCode":404,"retries":0}, incidentId: a895da35-6220-43eb-8cbd-f973626faa78 ErrorWithoutCodeError: No error message given, in payload ''
    at /home/jhford/taskcluster/aws-provisioner/node_modules/fast-azure-storage/lib/blob.js:425:27
    at tryCallOne (/home/jhford/taskcluster/aws-provisioner/node_modules/promise/lib/core.js:37:12)
    at /home/jhford/taskcluster/aws-provisioner/node_modules/promise/lib/core.js:123:15
    at flush (/home/jhford/taskcluster/aws-provisioner/node_modules/asap/raw.js:50:29)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9) +1ms
```

We should figure out what's going wrong here.

@owlishDeveloper: I'm sorry that I was unable to merge your patch, it bitrotted and I wasn't able to apply it cleanly.  I did use it to know where to update, but I was hoping you might be able to double check that I did everything needed to switch to the new `ping()` api


owlishDeveloper:
Everything looks good except that one comment I have :)

src/secret.js
```diff
@@ -1,17 +1,17 @@
-let base = require('taskcluster-base');
+let Entity = require('azure-entities');
```
elenasolomon: 'azure-entities' is used, but is not defined in package.json as a dependency

src/exchanges.js
```diff
@@ -1,7 +1,7 @@
-let base = require('taskcluster-base');
+let Exchanges = require('pulse-publisher');
```
elenasolomon: 'pulse-publisher' is used, but not defined in package.json as a dependency

src/api-v1.js
```diff
@@ -1,5 +1,5 @@
 let log = require('./log');
-let base = require('taskcluster-base');
+let API = require('taskcluster-lib-api');
```
elenasolomon: 'taskcluster-lib-api' is used but not defined in package.json as a dependency

jhford:
@owlishDeveloper thanks for taking a look!  I've verified that the `ping` method is removed.

@elenasolomon I will look at deploying this tomorrow first thing in the morning!

PR #138, opened 2017-04-03
bug 1318648 - support instance instantiation in us-east-2

grenade:
in anticipation of increased load for g2.2xlarge (gecko-t-win7-32-gpu, gecko-t-win10-64-gpu worker types) enable instance instantiation in us-east-2 to mitigate cost and availability issues

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -17,6 +17,7 @@ defaults:
       - 'us-west-1'
       - 'us-west-2'
       - 'us-east-1'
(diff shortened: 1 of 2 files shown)


jhford:
the patch should be all that's needed in the provisioner if I remember correctly, but will also need to make sure the following is addressed:

* security groups
* vpc configs
* cloud-mirror (new and old instances)

PR #139, opened 2017-04-13
New state gathering

jhford:
Based on feedback from AWS we should be doing state updates a little differently.


jhford:
So I'm seeing this working on the staging provisioner app just fine, but i'm very hesitant to promote it to production because I'm about to go on a week of PTO.  @jonasfj would you be OK handling a rollback if needed?

jonasfj:
Sorry, I just saw this hidden in emails from github about rfcs... :) I'll take a look tomorrow... Enjoy you PTO..

package.json
```diff
@@ -21,46 +21,46 @@
...
   },
   "dependencies": {
-    "assume": "^1.2.5",
-    "aws-sdk": "^2.23.0",
-    "azure-blob-storage": "^1.0.2",
-    "azure-entities": "^1.2.1",
-    "azure-storage": "^1.0.1",
+    "assume": "^1.5.0",
+    "aws-sdk": "^2.41.0",
+    "azure-blob-storage-temp-fork": "^1.0.4",
```
jonasfj: Not sure what you're doing here... reference a github branch is a nicer hack...

jonasfj: Not that there is any reason to change it now... Just a general recommendation... Pollution of npm is not a nice thing to do...

src/aws-manager.js
```diff
@@ -24,6 +24,102 @@ function dateForInflux(thingy) {
...
+ * This method will run an EC2 operation.  Because the AWS-SDK client is so
+ * much fun to work with, we need to do the following things above what it does
+ * to get useful information out of it.
+ *
+ *   1. We want to have exceptions that always have region, method and service
+ *      name if available
+ *   2. Any requests which would have a requestId should include it in their
+ *      exceptions
+ *   3. Sometimes promises from AWS-SDK just magically never return and never
+ *      timeout even though we've set those options.  We have our own timeout
```
jonasfj: are we using the latest verison?

src/aws-manager.js
```diff
@@ -24,6 +24,102 @@ function dateForInflux(thingy) {
...
+    // set it to the values here.
+    if (!logObj.region) {
+      logObj.region = region;
+    }
+    if (!logObj.service) {
+      logObj.service = serviceId;
+    }
+
+    // We're going to add these in because they're handy to have
+    if (!err.region) {
```
jonasfj: Wouldn't `object.Assign` or `_.defaults`...

src/aws-manager.js
```diff
@@ -24,6 +24,102 @@ function dateForInflux(thingy) {
...
+  } catch (err) {
+    let logObj = {
+      //err,
+      method,
+    };
+
+    // We want to have properties we think might be in the error and
+    // are relevant right here
+    for (let prop of ['code', 'region', 'service', 'requestId']) {
+      if (err[prop]) { logObj[prop] = err[prop]; }
```
jonasfj: `_.defaults(logObj, _.pick(err || {},  ['code', 'region', 'service', 'requestId']))`

src/aws-manager.js
```diff
@@ -24,6 +24,102 @@ function dateForInflux(thingy) {
...
+      if (logObj.requestId) {
+        logObj.requestIdFromHack = request.response.requestId;
+      } else {
+        logObj.requestId = request.response.requestId;
+      }
+    }
+
+    // For the region and service, if they don't already exist, we'll
+    // set it to the values here.
+    if (!logObj.region) {
```
jonasfj: Wouldn't `object.Assign` or `_.defaults` would give more structure...

package.json
jhford: I'll delete it when it's sorted out.

src/aws-manager.js
jhford: yes.  The `aws-sdk` is just not that good in my opinion, this is an example.  This is why I lean more towards using `aws4`; i don't trust `aws-sdk` to always actually work.

jhford:
Adding better error information as well as removing all concurrency from the describe* calls, and doing multiple describe* calls using filters so we're making smaller API calls.

PR #140, opened 2017-04-25
[don't merge] Describe by worker type

jhford:
Just opening this to track the work

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -242,10 +242,111 @@ class AwsManager {
     this.reportAmiUsage = series.amiUsage.reporter(influx);
   }
 
+  /** 
+   * Get the instances and spot requests for a given worker type.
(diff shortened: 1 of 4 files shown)


src/aws-manager.js
```diff
@@ -242,10 +242,111 @@ class AwsManager {
...
+      workerTypes = [workerTypes];
+    }
+    assert(Array.isArray(workerTypes));
+
+    let instances = [];
+    let requests = [];
+    let deadInstances = [];
+    let deadRequests = [];
+
+    let keyPrefixes = workerTypes.map(workerType => negate ? '!' : '' + this.keyPrefix + workerType);
```
jhford: This needs to be changed so that when negating, we search for a `!` of all the known workerTypes, but that we do a positive filter for `this.keyPrefix + '*'`.

PR #142, opened 2017-05-09
Stop using Influx

jhford:
Rather than changing the code too much right now, let's just use a NullDrain for lib-stats.  Because of how the metrics are reported, actually removing the calls might be a little more difficult

### DIFF
--- a/src/influx-series.js
+++ b/src/influx-series.js
@@ -1,4 +1,3 @@
-let stats = require('taskcluster-lib-stats');
 
 // This is a time series to measure how long it takes for instances to show up
 // in the AWS api responses
--- a/src/main.js
+++ b/src/main.js
@@ -275,16 +275,7 @@ let load = loader({
   influx: {
     requires: ['cfg'],
     setup: ({cfg}) => {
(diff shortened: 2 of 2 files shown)


jhford:
Let me know if this looks sane to you... if so, let's deploy this and kill our influx account.

jhford:
The biaser has always been best effort, and fully wrapped in a try/catch/drop-on-the-floor.

gregarndt:
Looks like these changes still keep references around for influx, but just turns it into a null drain.  What if we killed all references to influx as well and where we needed to send some stats, we use the monitor package to send to signalfx?

jhford:
we could do that, but because we pass positional arguments with the
influx drain around, it's a much riskier change
than using the null drain.  I wanted to turn it off as quickly as I
could, but I'd like to save the large work that'll be required to
remove the influx references for the EC2 Manager integration that's
going to happen.  That'll be a pretty large amount of work which will
cover the same parts of the code, so I'd like to do it together.

gregarndt:
Ok, didn't realize that killing it completely will be part of the ec2 manager stuff....good times 👍

PR #145, opened 2017-06-06
Move delay intervals to env variables

gregarndt:
At times we need to adjust the delays between iterations and describe (spot) instance requests.  instead of submitting it via code commits, let's have the dials in the app settings.

Note this will require setting the following app settings:
ITERATION_INTERVAL
DESCRIBE_INSTANCE_DELAY
DESCRIBE_SPOT_INSTANCE_DELAY

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -186,7 +186,7 @@ async function runAWSRequest(service, method, body) {
  *      submit data points to a statsum instance
  */
 class AwsManager {
-  constructor(ec2, provisionerId, keyPrefix, pubKey, maxInstanceLife, influx, monitor, ec2manager) {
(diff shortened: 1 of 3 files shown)


src/aws-manager.js
```diff
@@ -349,7 +351,7 @@ class AwsManager {
             },
           ],
         });
-        await delayer(1000)();
+        await delayer(describeSpotInstanceDelay)();
```
jhford: These are describing spot requests, not spot instances.  It's probably minor, but it might be confusing to use this name

src/aws-manager.js
```diff
@@ -194,6 +194,8 @@ class AwsManager {
     assert(maxInstanceLife);
     assert(influx);
     assert(monitor);
+    assert(describeInstanceDelay);
+    assert(describeSpotInstanceDelay);
```
jhford: I might be missing something, but I'm not really sure where these variables are being declared

gregarndt:
I believe I addressed the review comments, please take another look, thanks :)

PR #150, opened 2017-06-20
Provisioner using ec2-manager

jhford:
Using EC2 Manager to take over provisioning.  I've tested that this does launch instances.  This patch leaves a couple of EC2 related things in the Provisioner, mainly around pricing and launch specification generation.  In time, my idea is to move *all* ec2 related things into the `ec2-manager`, but I wanted the first pass to be something concrete without having to reinvent the wheel.

Things like the zombie killer and rogue killer will need to be implemented still, but are not a blocker.


src/worker-type.js
```diff
@@ -371,6 +371,7 @@ WorkerType.prototype.capacityOfType = function(instanceType) {
  */
 WorkerType.prototype.createLaunchSpec = function(bid) {
   assert(bid);
+  debugger;
```
imbstack: `debugger` be gone!

PR #151, opened 2017-06-26
stop sending the scopes list from the worker type definition

jhford:
While this list was not involved in any of the credential generation, we should still send the correct list of scopes back.

### DIFF
--- a/lib/api-v1.js
+++ b/lib/api-v1.js
@@ -496,8 +496,7 @@ api.declare({
   stability:  API.stability.stable,
   description: [
     'Insert a secret into the secret storage.  The supplied secrets will',
-    'be provided verbatime via `getSecret`, while the supplied scopes will',
-    'be converted into credentials by `getSecret`.',
+    'be provided verbatime via `getSecret`.',
     '',
     'This method is not ordinarily used in production; instead, the provisioner',
     'creates a new secret directly for each spot bid.',
@@ -515,7 +514,7 @@ api.declare({
       workerType: input.workerType,
       availabilityZones: [],
       secrets: input.secrets,
-      scopes: input.scopes,
+      scopes: [],
       expiration: new Date(input.expiration),
     });
   } catch (err) {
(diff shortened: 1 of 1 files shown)


djmitche:
Could we just remove the field altogether?

jhford:
We could, but then for a worker that wishes to verify which scopes it has, it would have to parse the certificate as json, which I thought was something we didn't want to do.

djmitche:
There's an API call for that (auth.currentScopes).

2017-07-04 5:57 GMT-04:00 John Ford <notifications@github.com>:

imbstack:
Do we have a way forward we want to go with this? I'm inclined to remove these fields altogether if they're not used, but am not a domain expert!

imbstack:
What do we want to do with this? Should we look into landing it again or close?

djmitche:
@jhford ^^

I'm in favor of removing.  The API method is marked as stable, but I think the only thing using it is the automation around updating generic-worker workerTypes.  I'd also be comfortable always returning an empty array and updating the docs to indicate that the field is unused, if that will prevent clients from breaking..

jhford:
@petemoore do you have any thoughts on whether we should just send an empty list or the actual list of scopes?

djmitche:
What about dropping the field entirely?  I don't think there are any clients that will break, including tools.

petemoore:
The generic-worker only uses [`credentials`](https://github.com/taskcluster/generic-worker/blob/56a6b31b7e91bf924e63a734662851684748499b/aws.go#L211-L213) and [`data`](https://github.com/taskcluster/generic-worker/blob/56a6b31b7e91bf924e63a734662851684748499b/aws.go#L243) properties, so from generic-worker perspective, it can be the full set of scopes, an empty set, or the field can be dropped entirely.

jhford:
@petemoore thanks for the update.

@djmitche I'd rather not remove the field entirely, but I've pushed a new version of the patch which stores and sends an empty list.  This should accomplish the goal of this PR with the absolute minimum risk of anything going wrong.

lib/api-v1.js
```diff
@@ -496,8 +496,7 @@ api.declare({
   stability:  API.stability.stable,
   description: [
     'Insert a secret into the secret storage.  The supplied secrets will',
-    'be provided verbatime via `getSecret`, while the supplied scopes will',
-    'be converted into credentials by `getSecret`.',
+    'be provided verbatime via `getSecret`.',
```
djmitche: *verbatim

djmitche:
WONTFIX'd

djmitche:
(in https://bugzilla.mozilla.org/show_bug.cgi?id=1375228)

PR #152, opened 2017-07-03
State directly from ec2 manager

jhford:
This also removes the scopes from the /state/ method because that information really isn't confidential.  This method wraps the EC2-Manager state and stats methods to provide a compatibility path forward until the tools website can be updated.  NOTE: The /state/ method's summary property is derived from a different call to the ec2-manager than the one to generate the lists of instances and requests.  This is done because it's non-trivial to use the result of the state call to ec2-manager in the summary generation function


src/api-v1.js
```diff
@@ -126,12 +127,8 @@ api.declare({
...
     let workerState;
     try {
-      workerState = await this.stateContainer.read(workerType.workerType);
-    } catch (err) {
-      if (err.code !== 'BlobNotFound') {
-        throw err;
-      }
-    }
+      workerState = await this.ec2manager.workerTypeStats(workerType.workerType);
+    } catch (err) { }
```
imbstack: Do we really want to throw away the error with no logging at all?

src/api-v1.js
```diff
@@ -722,76 +712,53 @@ api.declare({
...
-}, async function(req, res) {
-  let workerType;
-  let workerState;
-
-  if (!req.satisfies({workerType: req.params.workerType})) {
-    return;
-  }
-
+    workerState = await this.ec2manager.workerTypeState(workerType.workerType);
+  } catch (err) { }
```
imbstack: Same as above.

src/api-v1.js
```diff
@@ -722,76 +712,53 @@ api.declare({
...
-    workerType = await this.WorkerType.load({workerType: req.params.workerType});
-  } catch (err) {
-    if (err.code === 'ResourceNotFound') {
-      res.status(404).json({
-        message: req.params.workerType + ' does not have any state information',
-      }).end();
-    } else {
-      throw err;
+    workerStats = await this.ec2manager.workerTypeStats(workerType.workerType);
+  } catch (err) { }
```
imbstack: Same as above.

imbstack:
Looking good otherwise!

jhford:
I'll log those errors and deploy soon.

jhford:
Rebased and deployed with 21833d66b018b257a2bdfa75e4f7c03b9c3d4a69

PR #153, opened 2017-07-07
Worker Types should be created with no security groups

gregarndt:
I'm taking a guess that this is the right thing.  I hope it is.

Worker types should not need to specify a list of security groups in the
case where we are denying all inbound traffic.


gregarndt:
After speaking with John about this, I now realize that this PR is no longer needed for the purposes I intended which is to block all inbound traffic.  We will go the route of creating security groups in each of the regions that do not specify any inbound rules and allow all outbound traffic.

PR #157, opened 2017-08-10
Bug 1388436 - actually store availabilityZone data

djmitche:


jhford:
Looks good to me!  The confusion here for me is that I thought availability zones were being stored in the region objects.  Had that been the case, we'd have persisted the AZ info automatically.  I think keeping AZ as a sibling to region instead of a child is a good choice, though.

R+

djmitche:
Yeah, I think I made the same mistake writing it -- initially thought I'd put them in the regions, at which point updates to the data structure weren't necessary, and didn't reconsider that when I changed my mind.

PR #158, opened 2017-08-15
actually search worker.availabilityZones, fail if not found

djmitche:
This was looking at worker.AZs which is always undefined :(

I also added a check for an attempt to allocate in an unspecified AZ.  But will this work? I think the validation uses "fakezone1" which won't exist..

### DIFF
--- a/src/worker-type.js
+++ b/src/worker-type.js
@@ -466,7 +466,8 @@ WorkerType.createLaunchSpec = function(bid, worker, keyPrefix, provisionerId, pr
   // Find the AZ object, if any
   let selectedAZ = {};
   let foundAZ = false;
-  for (let r of worker.AZs || []) {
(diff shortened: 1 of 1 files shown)


src/worker-type.js
```diff
@@ -466,7 +466,8 @@ WorkerType.createLaunchSpec = function(bid, worker, keyPrefix, provisionerId, pr
   // Find the AZ object, if any
   let selectedAZ = {};
   let foundAZ = false;
-  for (let r of worker.AZs || []) {
+  let availabilityZones = worker.availabilityZones || [];
+  for (let r of availabilityZones) {
```
jhford: I know it's not new in this patch, but I was thinking that instead of r, maybe az would be a little more clear to read?

djmitche: Yeah, I'll make a PR for that too.

PR #160, opened 2017-08-18
Bug 1388436 - validation

djmitche:
In staging, this gets as far as validating against all AZ's, rather than using `fakezone1`.

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -333,7 +333,11 @@ class AwsManager {
       }
     } catch (err) {
       returnValue.canLaunch = false;
-      returnValue.reasons.push(err);
+      if (err.code === 'InvalidLaunchSpecifications') {
+        returnValue.reasons = err.reasons;
+      } else {
+        returnValue.reasons.push(err.toString());
+      }
       return returnValue;
     }
 
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -326,10 +326,11 @@ class AwsManager {
             this.provisionerId,
             'http://taskcluster.net/fake-provisioner-base-url',
             this.pubKey,
-            worker.workerType
+            worker.workerType,
+            this
         );
       } else {
-        launchSpecs = worker.testLaunchSpecs();
+        launchSpecs = worker.testLaunchSpecs(this);
       }
     } catch (err) {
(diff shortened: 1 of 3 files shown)


src/worker-type.js
```diff
@@ -414,14 +414,15 @@ WorkerType.prototype.createLaunchSpec = function(bid) {
  * function will throw if there is an error found or will return
  * a dictionary of all the launch specs!
  */
-WorkerType.prototype.testLaunchSpecs = function() {
+WorkerType.prototype.testLaunchSpecs = function(awsManager) {
```
jhford: I'd rather that we don't pass in `awsManager`.  This function is about whether or not the worker type config is structurally valid and not so much about whether those launch specs actually can or should be launchable.  Those checks should be done in `workerTypeCanLaunch` in the aws manager.  For this function, let's instead just assume that the list of available availability zones are those in the configuration, and if none are defined, just do it by region and instance type.  

If we absolutely have to have at least a single availability zone per region for the code to execute, let's instead just assume `${region}-a` with no per-az config.  Because of how availability zone naming works, I'm pretty sure that the 'a' zone is guaranteed to always exist.  Either way, since these launchSpecs are never actually used to create real instances, we should be fine to use a hypothetical availability zone.

src/aws-manager.js
```diff
@@ -326,65 +326,77 @@ class AwsManager {
...
+        returnValue.reasons = err.reasons;
+      } else {
+        returnValue.reasons.push(err.toString());
+      }
       return returnValue;
     }
 
-    await Promise.all(worker.regions.map(async r => {
-      if (!this.ec2[r.region]) {
+    await Promise.all(Object.keys(launchSpecs).map(async r => {
```
jhford: I'd rather iterate over the data we expect instead of the data that we got.  Maybe we should add an assertion that the launchSpec at the expected region/zone/type actually is an object

src/worker-type.js
djmitche: OK, I'll move those checks.  I'll need to add more checks later for the SubnetId / SGId, so I'll put those in workerTypeCanLaunch, too.

djmitche:
I'm pursuing another option, so (assuming it works!) this won't be necessary.

PR #161, opened 2017-09-12
Bug 1388436 - support subnet IDs

djmitche:
This seems sufficient in my testing.  After our conversation today about requiring users to enter security group IDs instead of names, it all fell into place.  EC2 requires those group IDs in a different LaunchSpec field anyway, so there's no need to disambiguate old (default VPC) vs. new (explicit subnet / vpc) worker types.  No changes are required for EC2-manager, as it is just accepting the LaunchSpec as given.

This seems too simple, but it seems to work and I can't find anything missing.

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -333,7 +333,11 @@ class AwsManager {
       }
     } catch (err) {
       returnValue.canLaunch = false;
-      returnValue.reasons.push(err);
+      if (err.code === 'InvalidLaunchSpecifications') {
+        returnValue.reasons = returnValue.reasons.concat(err.reasons);
+      } else {
+        returnValue.reasons.push(err);
+      }
       return returnValue;
     }
 
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -374,20 +374,23 @@ class AwsManager {
         }
       }));
 
+      // check security group names
       let launchSpecsRegion = worker.instanceTypes.map(t => launchSpecs[r.region][t.instanceType].launchSpec);
-      let allSGForRegion = _.uniq(_.flatten(launchSpecsRegion.map(spec => spec.SecurityGroups)));
+      let allSGForRegion = _.uniq(_.flatten(launchSpecsRegion.map(spec => spec.SecurityGroups).filter(g => g)));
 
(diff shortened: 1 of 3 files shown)


djmitche:
The individual commits explain each change, if there's any question of purpose.

src/worker-type.js
```diff
@@ -675,7 +675,8 @@ WorkerType.testLaunchSpecs = function(worker, keyPrefix, provisionerId, provisio
           truePrice: 1,
           region: region,
           type: type,
-          zone: 'fakezone1',
+          // just test the 'a' AZ in each region..
```
jhford: We should only do this if there are no AZ in the worker type definition, otherwise we should do a test for each availability zone's specializations.

src/aws-manager.js
```diff
@@ -374,10 +374,11 @@ class AwsManager {
         }
       }));
 
+      // check security group names
       let launchSpecsRegion = worker.instanceTypes.map(t => launchSpecs[r.region][t.instanceType].launchSpec);
-      let allSGForRegion = _.uniq(_.flatten(launchSpecsRegion.map(spec => spec.SecurityGroups)));
+      let allSGForRegion = _.uniq(_.flatten(launchSpecsRegion.map(spec => spec.SecurityGroups).filter(g => g)));
 
-      let hasAllRequiredSG = await sgExists(this.ec2[r.region], allSGForRegion);
+      let hasAllRequiredSG = allSGForRegion.length === 0 || await sgExists(this.ec2[r.region], allSGForRegion);
```
jhford: I think it'd be a little more clear here if this was an if statement instead of an or

src/aws-manager.js
```diff
@@ -333,7 +333,11 @@ class AwsManager {
       }
     } catch (err) {
       returnValue.canLaunch = false;
-      returnValue.reasons.push(err);
+      if (err.code === 'InvalidLaunchSpecifications') {
+        returnValue.reasons = err.reasons;
+      } else {
+        returnValue.reasons.push(err.toString());
```
jhford: Let's serialize these when needed, in the api.  It should be something like 
```js
{reasons: launchInfo.reasons.map(x => x.toString()}
```

djmitche:
OK, I addressed those three comments.

djmitche:
tested in staging

PR #162, opened 2017-09-19
Bug 1400261 - consider only configured AZs

djmitche:
The schema change should not be a problem, as no production workeTypes have AZ's defined.

### DIFF
--- a/src/worker-type.js
+++ b/src/worker-type.js
@@ -824,14 +824,31 @@ WorkerType.prototype.determineSpotBids = function(managedRegions, pricing, chang
         .filter(r => _.includes(managedRegions, r.region))
         .map(r => r.region);
 
+    let configuredAzs = {};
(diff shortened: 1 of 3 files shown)


src/worker-type.js
```diff
@@ -824,14 +824,28 @@ WorkerType.prototype.determineSpotBids = function(managedRegions, pricing, chang
         .filter(r => _.includes(managedRegions, r.region))
         .map(r => r.region);
 
+    let configuredAzs = {};
+    this.availabilityZones.forEach(az =>
+      az.region in configuredAzs ?
```
jhford: An if statement would read a little more clearly here, would you mind switching to that?

djmitche: Hah, hassan and I were trying to find a one-liner for it instead of an if.  but I can change it back.

PR #164, opened 2017-10-12
Bug 1407037 - declare things from the provisioner

djmitche:


djmitche:
I have already added the necessary scopes for staging and production

config.yml
```diff
@@ -95,7 +99,9 @@ production:
 
 staging:
   app:
-    id: staging-aws
+    provisionerId: staging-aws
+    description: |
+      Staging instance for the AWS provisioner.  Not for production use.
```
helfi92: Extra space before "Not"

djmitche: `I     hate you   .` ;)

djmitche: hahah, even in monospace markdown strips consecutive spaces.  See, it's markdown, it doesn't matter.

src/main.js
```diff
@@ -119,6 +119,27 @@ let load = loader({
...
+        return;
+      }
+
+      await queue.declareProvisioner(cfg.app.provisionerId, {
+        stability: process.NODE_ENV === 'production' ? 'stable' : 'experimental',
+        expires: taskcluster.fromNow('36 hours'),
+        description: cfg.app.description,
+      });
+
+      await WorkerType.scan({}, {
```
jhford: This is unneeded.  We know that this call has happened when the worker type was created, so we shouldn't need to call this again.

src/api-v1.js
```diff
@@ -319,6 +322,9 @@ api.declare({
     workerType: workerType,
   });
 
+  let queue = new taskcluster.Queue({credentials: this.credentials});
```
jhford: In order to match the existing structure of the provisioner, `queue` should be a module in the main.js and passed in as context to the API.

src/worker-type.js
```diff
@@ -779,6 +780,18 @@ WorkerType.prototype.determineCapacityChange = function(runningCapacity, pending
   return capacityChange;
 };
 
+WorkerType.prototype.declareWorkerType = function({queue}) {
+  // keep the worker around well over 24 hours, as we will likely re-declare
+  // at least that often
+  const expires = taskcluster.fromNow('36 hours');
```
jhford: Why is there an expires field?  We derive that information from the worker type being seen or not in the queue.

src/worker-type.js
```diff
@@ -779,6 +780,18 @@ WorkerType.prototype.determineCapacityChange = function(runningCapacity, pending
...
+
+  return queue.declareWorkerType(this.provisionerId, this.workerType, {
+    stability: 'stable',
+    expires,
+    description: `${this.description}\n\nOwner: ${this.owner}`,
```
jhford: not blocking this review, but maybe an owner field would be merited in the queue method?

src/main.js
djmitche: The intent is to "refresh" this data periodically (here, roughly daily based on dyno cycling).  Similar to how we "refresh" the docs, schemas, and references.

src/worker-type.js
```diff
@@ -779,6 +780,18 @@ WorkerType.prototype.determineCapacityChange = function(runningCapacity, pending
   return capacityChange;
 };
 
+WorkerType.prototype.declareWorkerType = function({queue}) {
```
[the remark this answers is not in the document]
djmitche: Thanks -- I wasn't sure that was a better idea, but I'll do that.

src/worker-type.js
djmitche: Because the workerType might not be seen if no workers are provisioned.

src/worker-type.js
djmitche: Yeah, I was between that and removing the owner field from the provisioner's WorkerTypes table, and this seemed like a compromise :)

djmitche: (since none of the owner fields contain anything especially meaningful)

djmitche:
I think that accounts for all of the comments..

src/api-v1.js
djmitche: I passed it to WorkerType, at which point the API didn't need it anymore.

jhford: Did you mean to delete this line?  We don't use this queue instance any longer from what I can tell, but maybe I'm wrong.

src/main.js
```diff
@@ -379,8 +385,20 @@ let load = loader({
...
+    setup: async ({cfg, provisioner, server, queue, WorkerType}) => {
+      if (cfg.app.publishQueueMetadata) {
+        await queue.declareProvisioner(cfg.app.provisionerId, {
+          stability: cfg.app.stability,
+          expires: taskcluster.fromNow('36 hours'),
+          description: cfg.app.description,
+        });
+
+        await WorkerType.scan({}, {
+          handler: wt => wt.declareWorkerType(),
```
jhford: If this really is something which is wanted and it's supposed to be done at an interval, I'd rather express that by putting it in a taskcluster-lib-iterate and have it run periodically instead of relying exclusively on the 24-hour reboot.

djmitche: This follows the model of docs, schemas, and references.

Does lib-iterate guarantee an immediate call, or does it wait `waitTime` ms first?

djmitche: It does :) https://github.com/taskcluster/taskcluster-lib-iterate/pull/5

djmitche:
Updated

src/main.js
```diff
@@ -379,8 +385,33 @@ let load = loader({
...
+    setup: async ({cfg, provisioner, server, queue, WorkerType, monitor}) => {
+      if (cfg.app.publishQueueMetadata) {
+        const day = 24 * 60 * 60 * 1000;
+        let i = new Iterate({
+          maxIterationTime: day,
+          watchDog: day,
+          waitTime: day,
+          monitor,
+          handler: async (watchdog, state) => {
+            watchdog.stop();
```
jhford: any reason for this?  The watchdog will through in this case if the handler doesn't complete in 24hours.  I suspect if this code takes 24, something is drastically wrong.

src/main.js
```diff
@@ -379,8 +385,33 @@ let load = loader({
   },
 
   all: {
-    requires: ['provisioner', 'server'],
-    setup: async ({provisioner, server}) => {
+    requires: ['cfg', 'provisioner', 'server', 'queue', 'WorkerType', 'monitor'],
+    setup: async ({cfg, provisioner, server, queue, WorkerType, monitor}) => {
+      if (cfg.app.publishQueueMetadata) {
+        const day = 24 * 60 * 60 * 1000;
+        let i = new Iterate({
```
jhford: Can you move this to a top level component?

src/main.js
djmitche: I think that's OK -- the idea is that it doesn't really matter how much time this takes.

djmitche:
TC-lib-iterate's docs suggest it takes care of the error handling and reporting to sentry -- `maxFailures` suggests it catches a failure (rather than crash the process0, and `monitor` suggests that it will report that failure to sentry.  I will tune the `waitTimeAfterFail` to something shorter rather than wait 24h, though.

djmitche:
(continuing) Have I misunderstood your comment or how lib-iterate works?

djmitche:
I moved to a top-level component and removed the `watchdog.stop()`.  I didn't add `waitTimeAfterFail` after all, since tc-lib-iterate doesn't implement it.  I didn't change any of the error handling, for the reasons above.

PR #166, opened 2017-11-07
Add endpoint to kill an AWS worker

helfi92:


src/api-v1.js
```diff
@@ -414,6 +414,45 @@ api.declare({
 });
 
 api.declare({
+  method:     'delete',
+  route:      '/killWorker',
+  // actions with context=worker will a query of the form ?provisionerId=...&workerType=...&workerGroup=...&workerId=...
+  // https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/actions#context
+  query: {
```
jhford: Query strings are only for continuation token and limit, so these inputs shouldn't be provided here.

Please use route parameters instead.

src/api-v1.js
```diff
@@ -414,6 +414,45 @@ api.declare({
 });
 
 api.declare({
+  method:     'delete',
+  route:      '/killWorker',
```
jhford: this is not a REST endpoint.  This is also not the correct service to have this endpoint.

The correct URL to use is:

```
https://ec2-manager.taskcluster.net/v1/region/{WORKERGROUP}/instance/{WORKERID}
```

src/api-v1.js
```diff
@@ -414,6 +414,45 @@ api.declare({
...
+  // actions with context=worker will a query of the form ?provisionerId=...&workerType=...&workerGroup=...&workerId=...
+  // https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/actions#context
+  query: {
+    provisionerId: /./,
+    workerType: /./,
+    workerGroup: /./,
+    workerId: /./,
+  },
+  name:       'killWorker',
+  scopes: [['aws-provisioner:kill-worker:<provisionerId>/<workerType>/<workerGroup>/<workerId>']],
```
jhford: We should be using the `aws-provisioner:manage-worker-type:<workerType>` scope.

src/api-v1.js
```diff
@@ -414,6 +414,45 @@ api.declare({
...
+  input:      undefined,
+  output:     undefined,
+  title:      'Kill a Worker',
+  stability:  API.stability.experimental,
+  description: [
+    'Kill an AWS worker by supplying a query with its region (workerGroup) and instanceId (workerId).',
+  ].join('\n'),
+}, async function(req, res) {
+  const {workerGroup, workerId} = req.query;
```
jhford: Please switch this to route parameters.

src/api-v1.js
```diff
@@ -414,6 +414,45 @@ api.declare({
...
+
+  if (!workerGroup || !workerId) {
+    res.status(400).end();
```
jhford: please use `res.reportError()`

helfi92:
If we're taking the approach of path parameters instead of query, we won't be needing this api endpoint anymore. Closing this for now.

PR #167, opened 2017-11-07
Add kill action on startup

helfi92:


src/main.js
```diff
@@ -400,6 +400,13 @@ let load = loader({
               stability: cfg.app.stability,
               expires: taskcluster.fromNow('36 hours'),
               description: cfg.app.description,
+              actions: [{
+                name: 'kill',
+                title: 'Kill',
+                context: 'worker',
+                url: 'https://aws-provisioner.taskcluster.net/v1/killWorker',
```
jhford: This doesn't have the correct URL, it'll need to be a template based on

```
https://ec2-manager.taskcluster.net/v1/region/{WORKERGROUP}/instance/{WORKERID}
```

helfi92:
@jhford Switched the URL to be an ec2-manager one. Note that we will need to change https://github.com/taskcluster/ec2-manager/blob/master/lib/api.js#L407 to be of method 'POST' instead of `DELETE`.

src/main.js
```diff
@@ -400,6 +400,13 @@ let load = loader({
               stability: cfg.app.stability,
               expires: taskcluster.fromNow('36 hours'),
               description: cfg.app.description,
+              actions: [{
+                name: 'kill',
+                title: 'Kill',
+                context: 'worker',
+                url: 'https://ec2-manager.taskcluster.net/v1/region/:workerGroup/instance/:workerId',
```
djmitche: This should probably be a configuration option -- for example, if the staging provisioner talks to a staging ec2-manager, we would want a different URL here for the two environments.

djmitche: In fact, it's already available at `cfg.ec2Manager`.

helfi92: Good idea, thanks for pointing that out.

helfi92:
Opened https://github.com/taskcluster/ec2-manager/pull/42 to change the endpoint method from DELETE to POST.

src/main.js
```diff
@@ -400,6 +400,13 @@ let load = loader({
               stability: cfg.app.stability,
               expires: taskcluster.fromNow('36 hours'),
               description: cfg.app.description,
+              actions: [{
```
jhford: This needs a method field to specify which http method to use for the request.

jhford:
Fwiw, the format that I proposed in the original action task format https://github.com/taskcluster/taskcluster-rfcs/issues/88 would've dealt with this.  Maybe that proposal has some merit, especially with the new substitution...

helfi92:
Awesome. Thanks for the quick review. I will merge when everything else is ready.

PR #168, opened 2017-11-20
Only publish API references in production

jhford:


src/main.js
```diff
@@ -97,26 +97,29 @@ let load = loader({
...
-            name: 'api',
-            reference: v1.reference({baseUrl: cfg.server.publicUrl + '/v1'}),
-          }, {
-            name: 'events',
-            reference: reference,
-          },
-        ],
-      });
+    setup: async ({cfg, validator}) => {
+      if (cfg.app.publishMetaData) {
```
imbstack: Generally we do this in other services with `process.env.NODE_ENV === 'production` to avoid needing the extra config var, but either way is fine.

PR #169, opened 2017-11-20
Use 'terminate' instead of 'kill'

djmitche:
There's no need to use a violent term like "kill".  Terminate may remind
some of a series of action movies, but it's at least a little more
clinical, and anyway corresponds to the EC2 terminology.

This leaves the `name` unchanged, as it is still performing the same
action.  It just changes the displayed text.


djmitche:
I won't rush to deploy this -- it can go in next time other things get deployed.

PR #171, opened 2017-12-01
Bid randomly across all available regions / types / zones

djmitche:
Absent feedback from capacity-oversubscribed errors, this should help us
to find AZs that do have capacity, at the cost of bidding into more
expensive AZs.

This obviously isn't great, but it balances complexity against effectiveness at this, 4:30pm on a Friday.


djmitche:
verbal approval from jonas

PR #172, opened 2017-12-08
More robust pricing

jhford:
I think the issues we were seeing were related more to not having see a price change recently than prices being too low or not having available az.  I think this should address that issue by loading significantly more data (4h instead of 30m) as well as reading all values of the request instead of just the first page.  Both of those things worked well in highly volatile spot price markets, but not in the new world with smoother pricing.

As well, this also requests prices only for those instance types which the provisioner knows it might need to work with.  This is to lower load on the EC2 api and hopefully result in fewer API errors

### DIFF
--- a/src/worker-type.js
+++ b/src/worker-type.js
@@ -815,12 +815,6 @@ WorkerType.prototype.determineSpotBids = function(managedRegions, pricing, chang
 
   /* eslint-disable no-loop-func */
   while (change > 0) {
-    let cheapestType;
-    let cheapestRegion;
-    let cheapestZone;
-    let cheapestPrice;
-    let cheapestBid;
-
     // Utility Factors, by instance type
     let uf = {};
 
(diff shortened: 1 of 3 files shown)


jhford:
I think this goes a long way to fixing the issue.  I would like to double check the behaviour of the EC2 api around the describeSpotPriceHistory endpoint, but I suspect this will help us in the issue where no prices were found for a thing.  I think that was the underlying cause of the trouble last week -- but correct me if I'm wrong.

src/worker-type.js
```diff
@@ -897,22 +934,14 @@ WorkerType.prototype.determineSpotBids = function(managedRegions, pricing, chang
...
+    if (!cheapestBid) {
       log.error({
         workerType: this.workerType,
         priceTrace: priceTrace,
       }, 'could not create any bid');
       throw new Error('Could not create any bid for ' + this.workerType);
     }
 
-    // pick a random zone to bid into
-    let cheapest = _.sample(bidOptions);
```
jonasfj: So you go back to always pick the cheapest region/AZ,

won't that fail, because of capacity issues which is no longer reflected in the price.

Please see: https://bugzilla.mozilla.org/show_bug.cgi?id=1422184#c7

I don't see this handling the "capacity-oversubscribed" signal, which we talked about on irc at the time.

PR #173, opened 2018-01-09
Use the new Spot Model with runInstances

jhford:
This should be all we need to switch the provisioner to using the new spot model


jhford:
The bulk of the work is happening in https://github.com/taskcluster/ec2-manager/pull/49, fwiw

jhford:
@imbstack still OK with this landing?  It's running in the provisioner-staging app on heroku

PR #174, opened 2018-01-12
Bug 1428068: Update prices through ec2-manager

walac:
aws-provisioner has a duplicated logic from ec2-manager to retrive spot
prices from Amazon. We remove this duplicated code and get the prices
through ec2-manager service.

### DIFF
--- a/src/aws-manager.js
+++ b/src/aws-manager.js
@@ -212,13 +212,10 @@ class AwsManager {
    * Update the state from the AWS API
    */
   async update() {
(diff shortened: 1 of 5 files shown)


walac:
@jhford When you have some time, would you mind reviewing this?

jhford:
In light of the worker-manager RFC being adopted, let's implement this support directly in the ec2-provider for the worker-manager.  Thanks for the patch, it looks really good!

PR #175, opened 2018-01-19
Bug 1431750 - add additionalProperties to schemas where missing

petemoore:
Hi John,

The motivation for this PR is that we are always explicit about `additionalProperties` - which has the side effect that sometimes it is kind of obvious that it is `true` but we set it anyway. This change is a forerunner to making a separate change to making `additionalProperties` a mandatory field when `type: object` is specified. This change therefore makes everything conformant before we add the extra rigor.

Let me know what you think! 😄

### DIFF
--- a/schemas/create-worker-type-request.yml
+++ b/schemas/create-worker-type-request.yml
@@ -149,6 +149,7 @@ properties:
             - eu-central-1
         launchSpec:
           type: object
+          additionalProperties: true
           description: |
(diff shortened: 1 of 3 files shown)


petemoore:
On reflection, I realised we could just make `additionalProperties` mandatory, if `properties` is specified. Details in the bug. Pushed a new commit with the simpler changes!

PR #181, opened 2018-02-02
Report table metrics like other services

jonasfj:
I'm going to terraform some charts for these metrics :)

@jhford, I didn't test this... I see no reason it should cause problems, but....

See: https://github.com/taskcluster/taskcluster-queue/blob/master/src/main.js#L160


jhford:
closed with #188

PR #183, opened 2018-02-08
Bug 1436620 - install newer pulse-publisher

djmitche:
This should help avoid process crashes when the pulse socket is disconnected (https://github.com/taskcluster/pulse-publisher/pull/21)


djmitche:
deployed

PR #187, opened 2018-04-26
Moved common schema elements into shared schemas

petemoore:
This should reduce the amount of code duplication we currently have in generated taskcluster-client-go code. In [tcawsprovisioner package](https://godoc.org/github.com/taskcluster/taskcluster-client-go/tcawsprovisioner) we currently have:

```go
type GetWorkerTypeResponse
type GetWorkerTypeResponse1
type InstanceTypeConfiguration
type InstanceTypeConfiguration1
type LaunchSpec
type LaunchSpec1
type RegionConfiguration
type RegionConfiguration1
```

With this PR, we should reduce this to:

```go
type InstanceTypeConfiguration
type LaunchSpec
type RegionConfiguration
type WorkerTypeLastModified
type WorkerTypeResponse
```

### DIFF
--- a/schemas/availability-zone-configuration.yml
+++ b/schemas/availability-zone-configuration.yml
@@ -1,0 +1,32 @@
+$schema: http://json-schema.org/draft-06/schema#
+title: Availability Zone Configuration
+type: object
+description: Availability zone configuration
+properties:
+  availabilityZone:
+    type: string
+    description: |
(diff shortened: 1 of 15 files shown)


imbstack:
I think it was just part of spring cleaning. I don't quite recall why we did it in general though.

petemoore:
These changes were moved into #188.

PR #189, opened 2018-05-07
Update some dependencies

jhford:
a couple smaller dependency updates


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "heroku-prebuild": "echo $SOURCE_VERSION > .git-version",
     "lint": "eslint lib/*.js test/*.js --fix",
-    "unittest": "NODE_ENV=test node --stack-trace-limit=50 ./node_modules/mocha/bin/_mocha test/*_test.js",
+    "unittest": "NODE_ENV=test node --stack-trace-limit=50 ./node_modules/mocha/bin/_mocha --exit test/*_test.js",
```
petemoore: Nice spot! I see we are upgrading mocha from `^3.2.0` to `^5.1.1` so indeed this makes a lot of sense. ++

> The default behavior in v4.0.0 is --no-exit, where previously it was --exit.

lib/api-v1.js
```diff
@@ -506,7 +506,7 @@ api.declare({
   let input = req.body;
   let token = req.params.token;
 
-  await req.authorize({workerType: workerType});
+  await req.authorize({workerType: input.workerType});
```
petemoore: Nice fix! Did the test crash before, or fail/pass silently? Too bad JS isn't stricter to automatically spot things like this.

PR #193, opened 2018-09-20
Bug 1462768 - provide taskclusterRootUrl in userdata to workers

petemoore:
Hey John! I wasn't quite sure how to test this change, so it comes with the disclaimer it is in untested.

I think this shouldn't break any staging environment you might be using so long as you don't have bleeding-edge workers connected to it that use `taskclusterRootUrl` - but that might be a gotcha if you do want to do that - so just a warning!

Let me know if you'd like me to implement it differently.

Many thanks!

### DIFF
--- a/lib/worker-type.js
+++ b/lib/worker-type.js
@@ -605,19 +605,22 @@ WorkerType.createLaunchSpec = function(bid, worker, keyPrefix, provisionerId, pr
   let securityToken = slugid.v4();
 
   config.userData = {
-    data: config.userData,
+    availabilityZone: bid.zone,
     capacity: capacity,
-    workerType: worker.workerType,
+    data: config.userData,
+    instanceType: bid.type,
+    lastModified: worker.lastModified.toISOString(),
+    launchSpecGenerated: new Date().toISOString(),
+    price: bid.truePrice,
+    provisionerBaseUrl: provisionerBaseUrl,
     provisionerId: provisionerId,
     region: bid.region,
-    availabilityZone: bid.zone,
-    instanceType: bid.type,
-    spotBid: bid.price,
-    price: bid.truePrice,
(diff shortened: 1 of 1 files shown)


lib/worker-type.js
```diff
@@ -605,19 +605,22 @@ WorkerType.createLaunchSpec = function(bid, worker, keyPrefix, provisionerId, pr
...
+    launchSpecGenerated: new Date().toISOString(),
+    price: bid.truePrice,
     provisionerBaseUrl: provisionerBaseUrl,
+    provisionerId: provisionerId,
+    region: bid.region,
     securityToken: securityToken,
+    spotBid: bid.price,
+    // Bug 1462768 - pass in hardcoded value since aws-provisioner-v1
+    // won't support running in any cluster other than this one
+    taskclusterRootUrl: 'https://taskcluster.net',
```
djmitche: could we call this simply `rootUrl`?

petemoore: I thought about that, funnily enough, while I was writing it. It occurred to me that the provisioner passes on system information that pertains to both AWS and to taskcluster. Based on this, it felt appropriate to be explicit. Also so that it isn't confused with the former `provisionerBaseUrl` (which also includes "provisioner" for context). Lastly, we consistently use e.g. env variable TASKCLUSTER_ROOT_URL, so it felt sane to use the same naming convention. Using `rootUrl` would indeed be more concise, but I feel wouldn't really be more understandable. If this endpoint only provided taskcluster configuration, I think I'd be more in favour, but as it is, I prefer having the full explicit name.

jhford: I understand the desire for brevity here, but I think that Pete raises a good point.  The 'rooturl' term is used broadly, so while for us it's apparent that it's the taskcluster root url, I think it might lead to confusion.

PR #195, opened 2018-10-19
Bug 1499701 - upgrade azure to behave better on SAS errors

djmitche:
This is a major upgrade of azure-entities, in which the signature for
setup() changes.


djmitche:
Looks good in staging (successfully loaded the workerTypes, which exercises this change..)

PR #198, opened 2019-03-12
Bug 1534667 - stop using signalfx, sentry

djmitche:
I'll test this in staging -- at worst it will crash on startup, so that should be easy to spot :)


lib/api-v1.js
```diff
@@ -506,7 +506,7 @@ api.declare({
   let input = req.body;
   let token = req.params.token;
 
-  await req.authorize({workerType: workerType});
```
imbstack: Gah

PR #199, opened 2019-03-12
Bug 1534667 - stop using signalfx, sentry

djmitche:
(seeing if this runs better from the upstream repo)


djmitche:
nope, no checks, apparently.

PR #201, opened 2019-04-24
Bug 1375197 - prohibit secrets in worker-type configs

djmitche:
The schemas cover both input (create, update workertype) and output (get workertype).  I verified again that there are no secrets at present, so once this is in place we'll be "locked" in that situation.

John suggested long ago that removing fields from the workerType could cause issues.  I'm not really sure what those issues would be, but this PR is better-safe-than-sorry with a deprecated service.


djmitche:
Deployed successfully!

PR #206, opened 2019-08-09
Bug 1572761 - actually expire secrets

djmitche:


lib/api-v1.js
```diff
@@ -566,30 +567,31 @@ api.declare({
...
-      res.status(404).end();
-    } else {
-      throw err;
-    }
+  if (!secret || secret.expiration < new Date()) {
+    return res.reportError('ResourceNotFound', 'No such secret', {});
   }
+
+  // remove the token so that it can't be used repeatedly
+  await this.Secret.remove({
```
imbstack: Should we also be deleting those that are expired? Not a big deal either way.

djmitche: They won't work with this change, so I'm not going to mess with it more than necessary.

